PR #15, opened 2022-02-21
Parameter Scan Command

### LINKED ISSUE #6: Scan HGC ROC parameters
Goal is to add a new command in the DAQ menu that can scan over a range of parameter values.

Inputs: parameter name, min value, max value, step, n_events_per_step

Output: one raw data file with chunks of events n_events_per_step corresponding to each parameter value from min to max in step increments.

Skelpdar:
Adds a tool in the DAQ menu to scan over one individual parameter.

Closes #6

### DIFF
--- a/tool/pftool.cc
+++ b/tool/pftool.cc
@@ -785,6 +785,36 @@ void ldmx_daq( const std::string& cmd, PolarfireTarget* pft ) {
     }
     fclose(f);
   }
+  if (cmd=="SCAN"){
+    std::string pagename=BaseMenu::readline("Sub-block name :  ");
+    std::string valuename=BaseMenu::readline("Value name :  ");
+    int iroc=BaseMenu::readline_int("Which ROC :  ");
+    int minvalue=BaseMenu::readline_int("Minimum value :  ");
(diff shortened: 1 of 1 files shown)


tool/pftool.cc
```diff
@@ -785,6 +785,36 @@ void ldmx_daq( const std::string& cmd, PolarfireTarget* pft ) {
...
+      try{
+        pft->hcal->roc(iroc).applyParameter(pagename, valuename, value);
+
+        for (int ievt=0; ievt<nevents; ievt++) {
+          pft->backend->fc_sendL1A();
+          std::vector<uint32_t> event = pft->daqReadEvent();
+          fwrite(&(event[0]),sizeof(uint32_t),event.size(),f);
+        }
+
+      } catch (const pflib::Exception& e) {
```
tomeichlersmith, line 812: Do you want to continue attempting different parameters after one of them failed? I would have assumed that this `try-catch` would be outside the `for` loop, but maybe I'm not thinking of a use case.

tomeichlersmith (approved):
Looks good. All of my comments are minor and you can feel free to ignore them if you don't have the time.

PR #38, opened 2022-04-01
Fix -1 setting when setting bias on all connectors

Skelpdar:


tomeichlersmith (approved):
was able to compile, did not _actually_ test but we are in a busy time

Issue #47, opened 2022-04-04
Unified loading of several ROC-specific config files


Skelpdar:
To make it easy for shifters, there should be a single command to load config files on all ROCs that is ready for data taking.

That the pftoolrc points to one config file per board seems like a reasonable solution. I will attempt adding this.

tomeichlersmith:
Resolved with #48

PR #48, opened 2022-04-04
Specify board specific config files in pftoolrc

Skelpdar:
Solves issue #47  

One can now specify board specific config files in pftoolrc and load these with ROC->DEFAULT_PARAMS. This is so shifters won't have to manually load specific config files on different boards. 

Currently Rcfile options are not passed through to the menus, and I didn't know whether that was desirable, so I settled with making the PolarfireTarget instance save where the default config files lie.


tomeichlersmith (approved):
Good idea, this looks good

PR #51, opened 2022-04-04
single file to configure across the entire detector

tomeichlersmith:
Parsing of a parent yaml file for loading an entire configuration of the detector. This includes HGC ROC parameter settings as well as ad-hoc settings for other parts of the system (each of which must be added manually). Similar to the parameter settings YAML files, the keywords are case _in_-sensitive. There is additional special word `inherit` signalling that the child YAML nodes should be applied to all polarfires or hgcrocs on a polarfire (depending on the context).

## To Do
- [x] Implement application of parsed configuration
- [x] Document YAML configuration syntax
- [ ] Construct draft detector configuration to share with CERN folks


tomeichlersmith:
Not sure which roc files are "stable" from the POV of CERN folks, so I'm striking the last to-do item.

PR #52, opened 2022-04-06
Separate detector library

tomeichlersmith:
I'm not sure if this will be helpful so I'm leaving it off main for now. Basically, this just moves the detector library into its own submodule so that it can be its own cmake target and potentially used downstream in tb-online.


tomeichlersmith:
Deciding that this is not going to be helpful in the medium term, so I am scrapping it.

PR #60, opened 2022-04-09
Iss56 - Split PFtool into multiple TUs

EinarElen:
This PR splits most of pftool.cc into distinct translation units and adds a corresponding header for pftool itself. This makes quick development a lot easier, reduces risks of conflicts with multiple developers, and makes it easier to write simple functions that can be used in multiple places, and, most importantly, faster to understand for people that aren't familiar with the code. 

It's not the pretties way of doing it, e.g. all the TUs are called pftool_foo.h/cc, but its enough to get the benefits from it. I can of course update things if someone has a strong preference :) 

Currently, all menus are separated out from their corresponding TUs, but if https://github.com/LDMX-Software/pflib/tree/auto-load-menu gets added then that can be updated as well which would make updating things even easier.


tomeichlersmith:
I merged an updated `ALIGN` function by @Skelpdar in PR #62 just now, can you update this branch to include that new method?

EinarElen:
I was expecting to do as much :)

tomeichlersmith:
I just did it :+1: wasn't as hard as I was worried about

tomeichlersmith (approved):

EinarElen:
Nice!

PR #64, opened 2022-04-11
Auto load menu lines

tomeichlersmith:
still in development but this is _sweet_ and would open the door to creating python bindings to menu lines in a formulaic way. This is the _dream_ for folks who want to script processes interacting with pftool.


tomeichlersmith (commented):
Need general cleanup and more documentation

tomeichlersmith:
Manually copying developments to different branch to avoid removing other developments when trying to resolve conflicts.

PR #68, opened 2022-04-28
add rogue receiver which provides event data to pure virt func

tomeichlersmith:
different tasks can be done by implementing derived classes of this
DMAReadoutReceiver and then providing them to daq_dma_dest of the
RogueWBI


tomeichlersmith:
Closing as stale as we transition to ZCU-focused development. Re-opening can be done if this is applicable.

Issue #61, opened 2022-04-09
QOL features for pftool


EinarElen:
Some parts of working with pftool are a bit slow to use and could benefit from being a bit more automatable, even if they are less effort than e.g. the scans. 

- A general "run this before we get beam" task 
- Setting a parameter on all ROC halves to the same value 
- Setting the LED bias on all connections at the same time 
- probably more things to add that I can't quite remember at this moment 

Most of these can of course be done manually or, ideally, with a config file but when you are working on it directly, these are things that we have found to save us a bunch of time. I've written these features mostly already but in a rather hacky/on-demand way which isn't ideal and would diverge from main much too fast, so I'm making this issue to collect them and do it more properly.

EinarElen:
Some updates of whats in here as of 
b9d11fba3384136396f473f1c463e363575d8bcd 

- [x] Factor out certain parts of the code as functions so that they can be reused in other places without excessive quantities of copy/paste code snippets. See e.g. 865ab7839b0b05b91f30c2274ba28854ad6401a4
- [x] Add versions of such functions that don't ask the user for input. See e.g. f2f8ee0ded12a0d3d4f4873931de0c173cdbe3c4 
- [x] Moved some function static variables, hardcoded constants, or variables that were being requested multiple times into static variables in wrapper functions. These are guaranteed to be initialized the first time the function is called which means the user will only be asked for input once throughout the duration of the program. It's still a global variable even if its protected by the scope of the function though... Useful for things that aren't likely to need changing within one session (like the number of channels per elink), although if you would need to you could always just write the BaseMenu::readline_foo call instead of using these. See e.g. 2ce646421a6e2fa881e44f72045e369cea9fe4c7
- [x] Generally, factor out code that takes up a lot of real estate (LOC or mental) that aren't directly relevant to the functions that they are part of. Typically, this is things like sprintfing together a filename in a buffer in the middle of a function, see 478f8ea1bc3015bba28f2201728036769907ffc4
- [x] Add versions of features like POKE_PARAM that are applied to all halves of the connected ROCs and/or all channels of the connected ROCs. Note: There is a slight difference in how you use these. Instead of providing a page like DIGITAL_HALF_0, you only provide the template part (e.g. DIGITAL_HALF_). This isn't super intuitive, but its enough for what we need right now
- [x] Add TOA to the csv output of the charge injection scans 
[...]

PR #70, opened 2022-04-28
Add TOA scan

cmantill:
Added TOA scan working off from [#61](https://github.com/LDMX-Software/pflib/issues/61). I am not sure how much it has diverged and would like to consult w. @jmuse13 to integrate his scans here as well before this is merged.


tomeichlersmith:
I merged in Joe's draft TOT scan with #71 so now we have to figure out how to resolve those conflicts.

PR #76, opened 2022-05-11
Auto load menu

tomeichlersmith:
Same rationale as #64 just updated to post TB developments.

### DIFF
--- a/tool/Menu.h
+++ b/tool/Menu.h
@@ -9,6 +9,7 @@
 #include <vector>
 #include <functional>
 #include <iostream>
+#include <iomanip>
 
 #ifndef PFLIB_TEST_MENU
 #include "pflib/Exception.h"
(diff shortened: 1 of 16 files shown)


tomeichlersmith (commented):
Looks good from my POV. Basically removed the need to manually list the menu lines in another file. Each menu can list its own lines in its own file and then it is propagated to `pftool` at library load time. This makes the actual executable pftool.cc much more stable and headers for the different TUs are only necessary to share information between separate TUs (not with pftool.cc).

EinarElen:
I think this looks really good. I haven't had a chance to run myself but I can't see any potential issues

EinarElen (commented):
I'm not 100% sure about removing the headers for each TU if they aren't explicitly used for communication between TUs. One of the advantages of having headers is to be able to browse the interface of a TU quickly without having to go through the implementation details

PR #80, opened 2022-05-30
Iss79 bias for active boards

EinarElen:
This fixes iss79 and does a small refactoring to the pftool_bias TU, renaming the `set_bias_on_all_connectors` function to `set_bias_on_all_boards` which is a more accurate description of what it actually does. Then goes ahead and factors out the code that actually does set the bias on each connector of a board into a function with a similar (but not the same) name.


tomeichlersmith (approved):
Looks good from my POV

PR #93, opened 2025-02-19
add multiple tries when reading bytes over I2C

tomeichlersmith:
In the HGCROCV3b, there is an imperfection in the I2C block which may occasionally result in I2C errors -- there is also the issue of things just locking up, which is what the fw-loader reload addresses.

## To Do
- [ ] make sure this functions


tomeichlersmith (commented):
This is functioning as best as I can tell.

PR #106, opened 2025-04-23
don't think extra check is necessary

tomeichlersmith:
only helpful in case where input file is broken but then the output
numbers would be wack and there would log messages that could be enabled
for investigation

### DIFF
--- a/app/pfdecoder.cxx
+++ b/app/pfdecoder.cxx
@@ -122,7 +122,7 @@ int main(int argc, char* argv[]) {
   // the output CSV if desired
   int count{0};
 
-  while (r and not r.eof()) {
+  while (r) {
     pflib_log(info) << "popping " << count << " event from stream";
     r >> ep;
     pflib_log(debug) << "r.eof(): " << std::boolalpha << r.eof() << " and bool(r): " << bool(r);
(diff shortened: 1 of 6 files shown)


tomeichlersmith:
Alright, this naive approach still failed because I forgot that `bool(r)` does not check for `eof`. I kept them separate in the implementation of `Reader.h` and `FileReader.h`.

I have a file with 98 events in it.
```
eichl008@cmslab1 ~/ldmx/pflib/build> hexdump -v -e '1/4 "%08x" "\n"' ../data/jeremy.raw | grep -c 'beef2025'
98
```
And now when running without a `-n` limit, we get 98 events in the output CSV file. This corresponds to 98*74+1 = 7253 rows since there is a header row and 74 rows per event.
```
eichl008@cmslab1 ~/ldmx/pflib/build> wc -l ../data/jeremy.csv 
7253 ../data/jeremy.csv
```
This at least eliminates the possibility of `pfdecoder` causing this off-by-one error.

tomeichlersmith:
I just double checked and it doesn't appear like we are losing any events **on the software side here** when running `pftool -> DAQ -> PEDESTAL`.

Maybe in certain run conditions, the first 1-2 events are empty and then discarded due to some other issue? I'm unsure, but I'm going to merge this and close the issue for now.

Issue #108, opened 2025-05-15
Charge Injections and LED Flashing in pftool


Skelpdar:
Sending commands for internal and external charge injections seem to be implemented in the fast control already:
https://github.com/LDMX-Software/pflib/blob/3a94560a4242377bc9a7bbd0990a9b71cc846fb9/src/pflib/FastControlCMS_MMap.cxx#L199 

But, we can not make use of these in pftool just yet. Assuming that these two FC commands work correctly, this is just a matter of implementing some menu options on the user interface side.

Features:
- Add a DAQ->CHARGE_INT menu option (simple)
- Add a DAQ->CHARGE_EXT menu option (simple)
- Add functionality to control LED bias voltages from pftool (Lennart has some bash scripts to do this, so follow his recipe), e.g. a BIAS->LED_BIAS menu option

tomeichlersmith:
The code is there to request a charge pulse, but we need to figure out the input parameters for this pulse. Specifically, the time difference between when we send the charge pulse request and when we send a L1A.

We can determine this time difference by recording the "pipeline depth of the ROC" and comparing this value to a fast control parameter that determines this time difference. The relationship between the pipeline depth and this time offset should be pretty stable, so once we scan the time offset parameter while keeping the pipeline depth fixed, we should be able to code that offset into pflib.

From looking at the HGCROC documents I have available to me, I can't figure out what this pipeline depth or the time offset parameters correspond to, but I wanted to keep my notes out in the open in case someone else can add to them.

tomeichlersmith:
And to be even more sure, I did a run with 100 events per `FAST_CONTROL.CALIB` setting and it confirms that we see the pulse at a setting of 20.

![Image](https://github.com/user-attachments/assets/a8622469-ae33-4359-bd09-436025ce558e)

verenamh:
As Andreas mentioned previously in this issue, we have now also tried testing the parameter_timescan function with more than one event per timepoint, we chose 50 events here. We have so far only done this with the lowrange capacitor and I have now additionally implemented a heatmap plotting option into parameter-time-scan.py using numpy.histogram2d (I will wait with pushing it to github until Andreas' code is approved into the main branch). The normalization was done by slicing the time axis into the different timepoints and normalizing these slices to a scale from 0 to 1. For the x_bins, I chose the amount of timepoints and the y_bins I chose 128 (from 1024 : 8 = 128 for 1024 ADC bins)

[...]

PR #121, opened 2025-05-30
Enable and time-align internal charge injection

tomeichlersmith:
A new `DAQ.DEBUG` command is present that helped me figure out the timing of the L1A relative to the calibration pulse in order to be able to observe it. (See https://github.com/LDMX-Software/pflib/issues/108#issuecomment-2922679386). This led me to update the default setting for the L1A following FastControl command.

In addition, I've updated `daq_run` to at least _call_ `fc.ledpulse()` if `cmd="LED"`. I'm hoping that we can time-align the LED pulse in the same way I did the charge pulse just with a different `fc_setup_*` function than the one currently in `CHARGE_TIMEIN`.

### DIFF
--- a/app/tool/pftool.cc
+++ b/app/tool/pftool.cc
@@ -790,7 +790,7 @@ class DecodeAndWriteToCSV {
  * Do a DAQ run using the input target
  *
  * @param[in] pft target to use
- * @param[in] Request function that does the appropriate request given a FastControl object
+ * @param[in] cmd PEDESTAL, CHARGE, or LED (type of trigger to send)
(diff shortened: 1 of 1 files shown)


tomeichlersmith:
With this branch, I can use `ROC.POKE` to set the parameters listed in the image. I then do a `DAQ.PEDESTAL` run to get the pedestals and then a `DAQ.CHARGE` run. The pedestals are the means of the ADC values in each `link_channel` from the `DAQ.PEDESTAL` run which are then subtracted from the ADC values of the `DAQ.CHARGE` run.

charge

PR #157, opened 2025-06-25
Added heatmap plotting option

verenamh:
I added an option to create 2D plots that show the number of counts recorded in each ADC-time bin (“heatmap plots”) for charge injection pulses with different CALIBs that were recorded with multiple events per timepoint. Also includes a stacked plot, to plot all pulses in one plot, for better comparison. It now needs to be chosen whether the user wants a SCATTER or HEATMAP plot when running the plotting script.


tomeichlersmith (approved):
Awesome! Thank you :)

PR #161, opened 2025-06-27
Pedestal Trimming

OwenCole04:
Added scans for parameters trim_inv, inv_vref, and noinv_vref in tasks. Also created a python script in ana to analyze results of trim_inv scan.

### DIFF
--- a/app/tool/tasks.cxx
+++ b/app/tool/tasks.cxx
@@ -83,26 +83,46 @@ load_parameter_points(const std::string& filepath) {
  */
 static void charge_timescan(Target* tgt) {
   int nevents = pftool::readline_int("How many events per time point? ", 1);
-  bool preCC = pftool::readline_bool("Use pre-CC charge injection? ", false);
-  bool highrange = false;
-  if (!preCC) highrange = pftool::readline_bool("Use highrange (Y) or lowrange (N)? ", false);
-  int calib = pftool::readline_int("Setting for calib pulse amplitude? ", highrange ? 64 : 1024);
+  bool isLED = pftool::readline_bool("Flash LED instead of the internal calibration pulse?", true);
   int channel = pftool::readline_int("Channel to pulse into? ", 61);
+  int link = (channel / 36);
+  auto channel_page = pflib::utility::string_format("CH_%d", channel);
+  auto refvol_page = pflib::utility::string_format("REFERENCEVOLTAGE_%d", link);
   int start_bx = pftool::readline_int("Starting BX? ", -1);
   int n_bx = pftool::readline_int("Number of BX? ", 3);
-  std::string fname = pftool::readline_path("charge-time-scan", ".csv");
   pflib::ROC roc{tgt->hcal().roc(pftool::state.iroc, pftool::state.type_version())};
-  auto channel_page = pflib::utility::string_format("CH_%d", channel);
-  int link = (channel / 36);
-  auto refvol_page = pflib::utility::string_format("REFERENCEVOLTAGE_%d", link);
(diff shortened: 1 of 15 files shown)


OwenCole04:
Also, Taylor and I were working on the same repo initially, so the commit history is a bit muddied. We split it up a few days ago and I only kept the additions I made in here.

app/tool/tasks.cxx
```diff
@@ -252,6 +252,191 @@ static void gen_scan(Target* tgt) {
...
+      .apply();
+      header["noinv_vref"] = noinv_vref;
+      
+      //do adc pedestal run for each channel; definitley not coded correctly but unsure how to rewrite
+      //scan to take scan over all channels at same time (since i'm only changing a link-wide parameter)
+      for (int ch = 0; ch < 72; ++ch){
+        header["channel"] = ch;
+    
+        tgt->daq_run("PEDESTAL", writer, nevents, pftool::state.daq_rate);
+      }
```
tomeichlersmith, line 433: Each `daq_run` includes all of the channels. The second lambda function input into the `writer` is where you decide which channels to write out (and how to write them out. In this case, it seems like you just want to do
```cpp
[&](std::ofstream& f, const pflib::packing::SingleROCEventPacket &ep) {
    f << ch << ',' << header["noinv_vref"].as_int64() << ',';
    ep.channel(ch).to_csv(f);
    f << '\n';
  }
}

File: app/tool/tasks.cxx [hunk 2]
```diff
@@ -252,6 +252,191 @@ static void gen_scan(Target* tgt) {
...
+      f << header["current_channel"].as_int64() << ',' << header["current_trim_inv"].as_int64() << ',';
+      ep.channel(header["current_channel"].as_int64()).to_csv(f);
+      f << '\n';
+    }
+  };
+
+  tgt->setup_run(1 /* dummy */, DAQ_FORMAT_SIMPLEROC, 1 /* dummy */);
+
+  for (int ch = ch_start; ch <= ch_end; ++ch) {
+    pflib_log(info) << "Running CH_" << ch;
```
tomeichlersmith, line 295: Similar to what I mentioned in the other comment. You don't need to do separate runs for the different channels unless you are worried about electronic cross talk. (We do not expect this to be an issue for pedestals I believe).

Instead of doing different runs, you would just apply all of the testing TRIM_INV to all the channels at once. e.g.
```cpp
auto trim_inv_test_builder = roc.testParameters();
  trim_inv_test_buildr.add("CH_"+std::to_string(ch), "TRIM_INV", trim);
}
auto trim_inv_test = trim_inv_test_builder.apply();
```

tomeichlersmith (commented):
This is good work! I am starting with some comments about how to avoid multiple runs when it isn't necessary.

I also want to point out that another student (Chelsea at UMN) is working on pedestal alignment https://github.com/LDMX-Software/pflib/issues/162
Maybe you can share your notes/comments/plots on that issue and y'all can collaborate to design an algorithm?

OwenCole04:
Hi Tom, I tried following through on the suggestions you made. Also, I'm going to post some of the procedure/plots I got from trim_inv_scan and subsequent pedestal runs to this issue. I got some plots for inv_vref scan and noinv_vref scan as well, but still unsure exactly what to look for.

tomeichlersmith (approved):
Awesome! This looks great :) thank you

OwenCole04:
Just a quick rundown on the trim_inv scan
[pedestal trimming.pdf](https://github.com/user-attachments/files/21126485/pedestal.trimming.pdf)

PR #169, opened 2025-07-03
Tot lund

andpet1324:
Script for analyzing TOT


andpet1324:
> Is there a reason to do a time scan while also scanning the ToA/ToT thresholds? I'm just worried about having one command that does everything. We can avoid errors by having many commands that do specific tasks.

When we're unsure when exactly the tot will trigger, yes. When we're confident on where to set the TOA threshold we shouldn't need it.

PR #175, opened 2025-07-09
167 toa alignment

taylorjcolaizzi:
Adding some quick scans in app/tool/tasks.cxx. These will be useful for doing a TOA alignment, since we'll want scans over the TOA parameters then. This pull request does not contain analysis scripts for those scans; it just provides a quick way to gather the data.

### DIFF
--- a/app/tool/tasks.cxx
+++ b/app/tool/tasks.cxx
@@ -642,6 +642,61 @@ static void sampling_phase_scan(Target* tgt) {
   }
 }
 
+/**
+ * TASKS.TOA_VREF_SCAN
+ *
+ * Scan TOA_VREF for both halves of the chip
+ * Check ADC Pedestals for each point
+ * Used to align TOA trigger to fire just above pedestals
+ * Should be done after pedestal alignment
+ */
+static void toa_vref_scan(Target* tgt) {
+  int nevents = pftool::readline_int("Number of events per point: ", 100);
+
+  std::string output_filepath = pftool::readline_path("toa_vref_scan", ".csv");
+
+  auto roc = tgt->hcal().roc(pftool::state.iroc, pftool::state.type_version());
+
+  int toa_vref = 0;
+
+  pflib::DecodeAndWriteToCSV writer{
+    output_filepath,
+    [&](std::ofstream& f) {
+      boost::json::object header;
+      header["scan_type"] = "CH_#.TOA_VREF sweep";
+      header["trigger"] = "PEDESTAL";
+      header["nevents_per_point"] = nevents;
+      f << "# " << boost::json::serialize(header) << "\n"
+        << "TOA_VREF";
+      for (int ch{0}; ch < 72; ch++) {
+        f << "," << ch;
+      }
+      f << "\n";
+    },
+    [&](std::ofstream& f, const pflib::packing::SingleROCEventPacket &ep) {
+      f << toa_vref;
+      for (int ch{0}; ch < 72; ch++) {
+        f << "," << ep.channel(ch).toa();
+      }
+      f << "\n";
+    }
+  };
+
+  tgt->setup_run(1 /* dummy */, DAQ_FORMAT_SIMPLEROC, 1 /* dummy */);
+
+  // Take a pedestal run on each parameter point
+  // Toa_vref has a 10 b range, but we're not going to need all of that
+  // since the pedestal variations are likely lower than ~300 adc.
+  for (toa_vref = 0; toa_vref < 256; toa_vref++) {
+    pflib_log(info) << "Running TOA_VREF = " << toa_vref;
+    auto toa_vref_test_builder = roc.testParameters();
+    toa_vref_test_builder.add("REFERENCEVOLTAGE_0", "TOA_VREF", toa_vref)
+    .add("REFERENCEVOLTAGE_1", "TOA_VREF", toa_vref);
+    auto toa_vref_test = toa_vref_test_builder.apply();
+    tgt->daq_run("PEDESTAL", writer, nevents, pftool::state.daq_rate);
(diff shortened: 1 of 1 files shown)


taylorjcolaizzi:
I ended up adding a short analysis script at the for observing the toa efficiency as a function of toa_vref. Generally, toa_vref should be set a little higher than the highest non-zero toa efficiency for each half of the chip.

app/tool/tasks.cxx
```diff
@@ -641,6 +641,128 @@ static void sampling_phase_scan(Target* tgt) {
...
+
+  // Take a pedestal run on each parameter point
+  // Toa_vref has a 10 b range, but we're not going to need all of that
+  // since the pedestal variations are likely lower than ~300 adc.
+  for (toa_vref = 0; toa_vref < 256; toa_vref++) {
+    pflib_log(info) << "Running TOA_VREF = " << toa_vref;
+    auto toa_vref_test_builder = roc.testParameters();
+    toa_vref_test_builder.add("REFERENCEVOLTAGE_0", "TOA_VREF", toa_vref)
+    .add("REFERENCEVOLTAGE_1", "TOA_VREF", toa_vref);
+    auto toa_vref_test = toa_vref_test_builder.apply();
```
tomeichlersmith, line 694: In this case, you don't need the intermediate `builder` variable
```suggestion
    auto toa_vref_test = roc.testParameters()
      .add("REFERENCEVOLTAGE_0", "TOA_VREF", toa_vref)
      .add("REFERENCEVOLTAGE_1", "TOA_VREF", toa_vref)
      .apply();
```

ana/toa_vref/toa_efficiency.py
```diff
@@ -0,0 +1,40 @@
...
+# requires a "toa_vref_scan.csv" file to run correctly.
+# Then, plots the efficiency versus toa_vref and saves the plot.
+
+import pandas as pd
+import matplotlib.pyplot as plt
+
+fp = 'toa_vref_scan.csv' # your file path here
+
+data = pd.read_csv(fp,
+                   skiprows = 1)
```
tomeichlersmith, line 11: See https://github.com/LDMX-Software/pflib/blob/7e9cec455051969671d18223bd8db990c1266228/ana/trim_inv/trim_inv_align.py#L9-L12

You can add a few lines of code before attempting to `from read import read_pflib_csv` and then you have access to the `read_pflib_csv` function which can do this `skiprows` for you and parse in the run parameters stored in the JSON header to (for example) include in the plot legend or title.

tomeichlersmith (changes requested):

taylorjcolaizzi:
Sounds good! I'm working on an analysis for the trim_toa_scan today, so hopefully I can push that up later and request a review!

ana/trim_toa/trim_toa_efficiency.py
```diff
@@ -0,0 +1,86 @@
...
+Ideally, we're looking for the calib value that
+corresponds to the start of non-zero toa_efficiency
+for each channel's trim_toa value.
+"""
+
+# separate the data into unique trim_toa values
+unique_trim_toas = data['TRIM_TOA'].unique()
+trim_lists = []
+for trim in unique_trim_toas:
+    trim_lists.append(data[data['TRIM_TOA'] == trim])
```
tomeichlersmith, line 63: I don't think this is used
```suggestion
```

app/tool/tasks.cxx
```diff
@@ -878,6 +878,134 @@ static void vt50_scan(Target* tgt) {
...
+    pflib_log(info) << "Running TRIM_TOA = " << trim_toa;
+    auto trim_toa_test_builder = roc.testParameters()
+    for (int ch{0}; ch < 72; ch++) {
+      trim_toa_test_builder.add("CH_"+std::to_string(ch), "TRIM_TOA", trim_toa);
+    }
+    for (calib = 0; calib < 4096; calib += 16) {
+      pflib_log(info) << "Running CALIB = " << calib;
+      // Set the TRIM_TOA and CALIB parameters for each channel
+      trim_toa_test_builder.add("REFERENCEVOLTAGE_0", "CALIB_2V5", calib).add("REFERENCEVOLTAGE_1", "CALIB_2V5", calib);
+      trim_toa_test = trim_toa_test_builder.apply();
```
tomeichlersmith, line 1003: Since `trim_toa_test_builder` is created outside of this `for` loop and you are `add`ing to it on each one, you are potentially combining parameters from multiple runs of the `for` loop. This isn't a problem for this specific case because you are using the same page and parameter names on each loop, but if (for example) you were changing the channel number of a parameter, it would break.

For this reason and to avoid future confusion, I am suggesting to move `auto trim_toa_test = trim_toa_test_builder.apply()` to before the `calib` `for` loop and then have a `auto calib_test` parameters applied within it.

```suggestion
    // set TRIM_TOA for each channel
    auto trim_toa_test = trim_toa_test_builder.apply();
    for (calib = 0; calib < 4096; calib += 16) {
      pflib_log(info) << "Running CALIB = " << calib;
      // Set the CALIB parameters for both halves
      auto calib_test = roc.testParameters()
        .add("REFERENCEVOLTAGE_0", "CALIB_2V5", calib)
        .add("REFERENCEVOLTAGE_1", "CALIB_2V5", calib)
        .apply();
```

tomeichlersmith (changes requested):
Looking good :) just a few more cleanup comments

app/tool/tasks.cxx
taylorjcolaizzi: Sounds good! I hadn't considered this complication before, so I'm glad you caught it!

tomeichlersmith (approved):

PR #178, opened 2025-07-11
Function to plot highrange correction for different channels to study ADC linearity

verenamh:
To further study the ADC linearity, I created a plotting function that plots both the lowrange and highrange data, then corrects the highrange data to align with the lowrange by doing linear fits of both. It requires two csv's as input, which is the lowrange and highrange data (in that order). It does this for multiple channels and stores the plot for each channel in a new folder called hrcorr. It also plots the lowrange and highrange slopes of the fits against the respective channel number to observe any differences channel-wise.

### DIFF
--- a/ana/charge/highrange_correction.py
+++ b/ana/charge/highrange_correction.py
@@ -1,0 +1,128 @@
+import update_path
+from pathlib import Path
+import argparse
+
+import matplotlib.pyplot as plt
+import numpy as np
+
+from read import read_pflib_csv
+from get_params import get_params
+
+parser = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


tomeichlersmith (changes requested):
This is a great idea!

I would ask for you to move this function to its own script. Its not really a "time scan" (x isn't time if I'm reading the code correctly), so it deserves its own script with its own name.
If you want to use functions from this script in your plotting function, that's good motivation to move them out into their own modules like `read.py` and `plt_gen.py` so that they can be shared with this script and yours.

tomeichlersmith:
I didn't see anything I would move into the shared modules, I'm just pointing out that _if there are_ functions you want to share, then those functions should go into the shared modules.

tomeichlersmith (approved):

PR #179, opened 2025-07-11
158 multi channel option for parameter timescan

verenamh:
Added a multi-channel option to the parameter timescan function, in order to study the ADC linearity for multiple channels simultaneously. It adds a column to the csv saving the channel number, and applies the same parameters to all 72 channels.

### DIFF
--- a/app/tool/tasks.cxx
+++ b/app/tool/tasks.cxx
@@ -485,24 +485,38 @@ static void parameter_timescan(Target* tgt) {
   if (totscan) toa_threshold = pftool::readline_int("Value for TOA threshold: ", 250);
   if (totscan) tot_threshold = pftool::readline_int("Value for TOT threshold: ", 500);
   int calib = pftool::readline_int("Setting for calib pulse amplitude? ", highrange ? 64 : 1024);
-  int channel = pftool::readline_int("Channel to pulse into? ", 61);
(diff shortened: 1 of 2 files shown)


app/tool/tasks.cxx
```diff
@@ -550,11 +567,14 @@ static void parameter_timescan(Target* tgt) {
     // Add implementation for other pages as well
     for (std::size_t i_param{0}; i_param < param_names.size(); i_param++) {
       test_param_builder.add(
-      (param_names[i_param].first == "REFERENCEVOLTAGE_0" ? refvol_page : 
-       param_names[i_param].first == "REFERENCEVOLTAGE_1" ? refvol_page :
-       param_names[i_param].first),
+      "REFERENCEVOLTAGE_0",
```
tomeichlersmith, line 570: This change breaks the ability for a parameter in the input parameter points file to be from a page that isn't `REFERENCEVOLTAGE` (`0` or `1`). Why did you make this change? What was the motivation?

If you need to only be changing the RV page parameters, then maybe you should make a new command instead of trying to make this (already pretty complicated) command work for your purposes?

verenamh:
I'd prefer it if you could remove the python changes to merge them separately :)

tomeichlersmith (approved):

PR #182, opened 2025-07-15
pedestal leveling

tomeichlersmith:
## To Do
- [ ] code cleanup and documentation, move task to own file with header
- [ ] add another iteration to level even more?

## Functionality Test
I ran this new task after a hard reset to check how it does. The pftool log of what I did is [180-commands.log](https://github.com/user-attachments/files/21240411/180-commands.log)
I then ran
```
$ python ana/pedestal/violinplot.py data/pedestal-2025-07-15-180927.csv -o default.png
$ python ana/pedestal/violinplot.py data/pedestal-2025-07-15-181017.csv -o leveled.png
```
to produce the following. The calib channels do not move (as expected, we don't bother leveling them).

default
leveled


taylorjcolaizzi:
> @erik-wallin @taylorjcolaizzi I am requesting your reviews to see how this functions on the HGCROCs you have. I want to make sure that it runs successfully. If you see results similar to mine that is an extra win.
> 
> I've been testing it by doing a `roc.hardreset` in order to see the default-parameter pedestals and then running `tasks.level_pedestals` in order to see the leveled pedestals. Note: after a `roc.hardreset` you will need to `daq.setup.standard`. The 180-check.log shows exactly what I did.
> 
> ```
> # on zcu
> cd pflib
> git switch 180-pedestal-leveling
> cd build
> make -j
> ```

Sure thing! I followed your steps and share my output here. I'm attaching a zip file of my yaml file, since Github won't let me upload .yaml here.

[level-pedestals-roc-0-settings_20250717_183223.yaml.zip](https://github.com/user-attachments/files/21302954/level-pedestals-roc-0-settings_20250717_183223.yaml.zip)
 
One thing I noticed is that `CH_42.TRIM_INV` = 2147483647, which might be a little high.

tomeichlersmith:
Thank you both! I now have two more things to look into.

- [ ] fix ultra-high TRIM_INV issue. I'm guessing this is because the algo is accidentally finding a negative TRIM_INV which is then being converted to a super-high value when it is static cast into a unsigned int.
- [ ] fix upper limit of TRIM_INV. two of the channels that are far below the link median (70 and 34) have a TRIM_INV bigger than 63 mean which is then cropped to only 6 bits and thus are not shifted close to the median at all.

taylorjcolaizzi:
[like]  Colaizzi, Taylor James (zxk4gs) reacted to your message:
________________________________

tomeichlersmith:
I'm going to merge this since it is behaving as expected and I can't think of anything else to do except look at updating the algorithm so that it performs better.

PR #189, opened 2025-07-28
Lund HGCROC configuration file after pedestal leveling

verenamh:
Added the configuration file for Lund to the config folder, to allow co-workers in Lund to use the same configuration in the future.


tomeichlersmith (approved):
Looks good!

PR #191, opened 2025-08-01
Separate commands in submenu into different files

tomeichlersmith:
This is mainly for developer ergonomics. When files start getting large, something starts scratching in the back of my brain and I want to factor the different parts out.

Separating out the commands into their own files helps by 
- keeping the files small and more manage-able
- forcing each command to explicitly list what it needs via includes
- addition/subtraction/movement of commands can happen in parallel and the only potential conflict is in the central file where they are registered with the menu

One might be concerned that these extra files could impact compilation time, and you'd be correct. We are trading compile-from-scratch times for iterative compilation times.
```
$ time just clean configure build
... omitting normal output
real    1m45.431s
user    0m0.404s
sys     0m0.137s
$ git switch main
$ time just clean configure build
... omitting normal output
real    1m24.924s
user    0m0.314s
sys     0m0.140s
```
Iterative compilation improves since we don't force the compiler to re-compile the other commands that we aren't editing.
```
# still on main
$ touch app/tool/tasks.cxx # mock edit a command
$ time just build
... omitting normal output
real    0m4.146s
user    0m0.116s
sys     0m0.108s
$ git switch - # back to this branch
$ just clean configure build
$ touch app/tool/tasks/parameter_timescan.cxx # mock edit a command
$ time just build
... omitting normal output
real    0m3.294s
user    0m0.162s
sys     0m0.059s
```


taylorjcolaizzi (approved):
I like having the different commands in their own files. That way, it's easier to search for individual commands instead of having to guess which large file they're hidden inside. Approve!

PR #197, opened 2025-08-04
link alignment via inv_vref

OwenCole04:
Modified inv_vref scan to only scan over one selected channel in each link. Added inv_vref_align.py in ana/inv_vref to take the results of this scan and find the optimal inv_vref value for each link to attain a target adc. (also changed my old trim_inv scan and algorithm to incorporate dacb similar to Chelsea's algorithm, although this isn't really pressing since we already have a method to level pedestals).


OwenCole04:
for reference: pedestal run with default inv_vref values after running level_pedestals on our ROC:

output

OwenCole04:
> This looks great :) the file diff doesn't look too bad so no reason to apologize.
> 
> Are you interested in converting this logic into C++ anytime soon? I think that is the next step so the tuning can be automated and run without having to create intermediate files.

Thanks! I would be interested in converting this to C++ soon, just realized though there's still one thing that might need some tweaking first though. 

Depending on what channels you select to scan, you can sometimes get weird behavior. For example, the plot above used inv_vref values using a scan of channels 17 and 51 (which I have set as the default for the inv_vref scan). Across multiple scans using these channels I didn't encounter anything off. I scanned a bunch of different channels on our roc, and some channels seemed to work consistently while others didn't. 

The pedestal run below used inv_vref settings from a scan of channels 23 and 54. In this case, channel 23 worked well and link0 hit its target adc of 100, while link1 using channel 54 didn't.

output

OwenCole04:
Thanks sounds good! I'll keep looking into it!

Looking back on a more comprehensive scan I did a while back with every single channel (ignore the red line), it looks like it can get quite messy sometimes 😅

image

image

PR #206, opened 2025-09-16
first draft for trim toa alignment

tomeichlersmith:
@taylorjcolaizzi worked on this over the summer and I'm just going to clean this up and merge it so we don't lose the work.

I'd rather have a poorly-working example to fix rather than start from scratch.


tomeichlersmith (commented):
- [ ] move support functions into their own headers/impl for sharing
- [ ] double-check AI slop
- [x] move `cout` into logging

PR #209, opened 2025-09-18
I2C interface cleanup 

erik-wallin:
By popular request I've 
- removed I2C::general_write_read_ioctl
- made it so I2C_Linux uses ioctl rather than file descriptor read-writes for everything. So ioctl is now used both for the ROC and bias communication
- cleaned up old testbeam stuff in Bias.cxx

I2C_Linux::read_byte and I2C_linux::write_byte now only makes one attempt at reading or writing and throws an exception immediately if it doesn't work.  

All tests are passing and this works well on the Lund test stand. But it would be nice if either @tomeichlersmith or @taylorjcolaizzi could try this out


taylorjcolaizzi:
Hi Erik! How would you recommend testing? Luckily, it compiles just fine on the UVA setup!

erik-wallin:
Hi @taylorjcolaizzi you could first run the tests ./test-pflib and ./test-pflib-on-zcu, and then maybe just try to read, write, and read back some ROC parameters to see if things make sense? That should be sufficient!

tomeichlersmith (approved):
While I'm unsure if we'll need to re-implement the multiple tries, I think we've observed stable-enough operation in the past when connecting directly to the HGCROC from the ZCU that I'm happy to drop the multi-tries.

At UMN, we are switching from direct HGCROC-ZCU connection to optical connection via the lpGBT (ZCU-lpGBT-HGCROC) that I still don't understand well enough to undo and test this myself. With that in mind, I think this looks good and if its working for you and @taylorjcolaizzi , I'm extremely confident it will work for our HGCROC-ZCU set up (if we ever go back to that).

PR #212, opened 2025-09-26
Add ECOND and ECONT I2C maps

cmantill:
The ECONs have a 16-bit I2C address (0x20 or 0x60 by default) with an 8-bit register space.

- Some parameters span across multiple registers and that is identified with `size_byte` in the yaml files.
- Some parameters only occupy part of the 8-bits and that is identified with the "param_mask" in the yaml files.
- The structure of the input yaml file is generally:
```
BLOCK:
  GLOBAL OR CHANNEL: 
    PARAMETER_NAME:
```

But some parameters are simply:
```
BLOCK:
  PARAMETER_NAME:
```

And some other parameters have double nested structure:
```
BLOCK:
  GLOBAL OR CHANNEL: 
    ANOTHER LEVEL:
      PARAMETER_NAME:
```

All of these cases are handled.

This PR compiles this information into header files.

Tested with:
```
python econ-to-header.py ECONT_I2C_params_regmap.yaml econt.h
python econ-to-header.py ECOND_I2C_params_regmap.yaml econd.h
```


CMakeLists.txt
```diff
@@ -33,6 +33,9 @@ endif()
 
 # Boost for CRC calculator, Test, and Logging
 find_package(Boost COMPONENTS log unit_test_framework REQUIRED)
+if(Boost_FOUND)
+    include_directories(${Boost_INCLUDE_DIRS})
+endif()
```
tomeichlersmith, line 38: Is this necessary? (I don't think it should be. We should be "linking" the Boost target to our targets which includes the necessary directories as well as links the libraries.)

register_maps/econ-to-header.py
```diff
@@ -0,0 +1,125 @@
...
+
+def safe_int(val):
+    """Convert YAML value to int whether it's a string like '0x0398' or already an int."""
+    if isinstance(val, str):
+        return int(val, 16) if val.startswith("0x") else int(val)
+    return int(val)
+
+def make_register_locations(address, mask, shift, size_byte):
+    """Split a register into 8-bit chunks if it spans multiple bytes.
+    syntax reminder: RegisterLocation(reg, min_bit, n_bits):
```
tomeichlersmith, line 17: Wouldn't it be 32-bit (4 byte) chunks because the registers are 32-bits wide?

The "RegisterLocation" developed for the HGCROC focuses on 8-bit wide registers since that's how big the registers are in the HGCROC, but I think the same struct will work with higher offsets/sizes for the larger registers on the ECOND.

tomeichlersmith (commented):
This does look good, I'm not certain when I'll sit down and give it a go myself, but this is a good start.

CMakeLists.txt
cmantill: Ah sorry this is a leftover from when I was trying to compile on my mac... Got rid of it

register_maps/econ-to-header.py
cmantill: No, no, let me clarify that the registers are also 8-bit wide (and some parameters span more than 1 byte).

tomeichlersmith:
My notes/thoughts as I'm looking at this...

- Are there versions of the ECON-D/T? We might want to specifically label this version in case we get a different version (with different parameters) in the future.
- I think I'm going to convert the `PageLUT` you wrote into a `ParameterLUT` just with a "page address" of 0 for all pages. This allows us to use the same types as used for the HGCROC and then the object that is deciding how to set the registers (in the HGCROC case, this is `ROC::setRegister` which combines the "page" and "register" into a full address) can ignore the "page address" and just use the "register address" as the full address.
- I still want to isolate the "compiler" into its own sub-library just because its complicated.
- How would you want to specify the ECON parameters in YAML? The HGCROC uses the same syntax as what is stored in the ParameterLUT: (i.e. "page" and "parameter" are both specified), but we could add more flexibility to the `extract` method. Perhaps doing the same interpretation you did in the python script in terms of converting a more-deeply nested set of parameters into a page/parameter name pair.

PR #228, opened 2025-10-30
Fix ECON slow control for register addresses that are split between multiple parameters

cmantill:
This PR solves two issues in the ECON slow control mapping

1. There were some missing registers in the .yaml to header conversion. These registers had an extra level of indentation in the yaml file that the python script did not account for.

2. The original header file included register locations (limited to 8 bits) where the bit shift was larger than 8 bits.
For example, for RegisterLocation(address, bit shift, mask):
```
{"GLOBAL_HEADER_MARKER", Parameter({RegisterLocation(0xf2c, 47, 8), RegisterLocation(0x0f2d, 0, 1)}, 486)},
```
this has now been changed to:
```
{"GLOBAL_HEADER_MARKER", Parameter({RegisterLocation(0x0f2c, 0, 0), RegisterLocation(0x0f2d, 0, 0), RegisterLocation(0x0f2e, 0, 0), RegisterLocation(0x0f2f, 0, 0), RegisterLocation(0x0f30, 0, 0), RegisterLocation(0x0f31, 7, 1), RegisterLocation(0x0f32, 0, 8)}, 486)},
```
so the register includes all possible register addresses and for those that are not used it sets the mask to 0.

This changed allowed to simplify the map of register address -> nbytes (in principle is not needed anymore but kept it for now).

Checks:
- I included these cases in the compile.cxx test.
- I also dumped the ECOND parameters in hardware after a hard reset and compared them with the default values. The mismatches correspond to status parameters in the ECOND which are truly different because of the initial state.

[compare_yaml.py](https://github.com/user-attachments/files/23240353/compare_yaml.py)
[econd_0_settings_20251030_160144.yaml](https://github.com/user-attachments/files/23240354/econd_0_settings_20251030_160144.yaml)
[ECOND_I2C_params_regmap.yaml](https://github.com/user-attachments/files/23240367/ECOND_I2C_params_regmap.yaml)

### DIFF
--- a/src/pflib/Compile.cxx
+++ b/src/pflib/Compile.cxx
@@ -181,8 +181,10 @@ std::map<uint16_t, size_t> Compiler::build_register_byte_lut() {
   // register address -> number of bytes used
   std::map<uint16_t, size_t> reg_byte_lut;
 
-  // assume that every parameter's RegisterLocation list already covers a
-  // contiguous byte range
+  /**
+   * @note assume that every parameter's RegisterLocation list already covers a
+   * contiguous byte range
+   */
   for (const auto& page_pair : page_lut_) {
     const std::string& page_name = page_pair.first;
     const Page& page = page_pair.second;


src/pflib/Compile.cxx
```diff
@@ -181,6 +181,8 @@ std::map<uint16_t, size_t> Compiler::build_register_byte_lut() {
   // register address -> number of bytes used
   std::map<uint16_t, size_t> reg_byte_lut;
 
+  // assume that every parameter's RegisterLocation list already covers a
+  // contiguous byte range
```
tomeichlersmith: Changing comment style to inject this into the docs

```suggestion
  /**
   * @note assume that every parameter's RegisterLocation list already covers a
   * contiguous byte range
   */
```

tomeichlersmith (approved):

PR #229, opened 2025-10-30
Add Data Capture from ECON

cmantill:
Starting a draft PR so that Charles can follow changes in DAQ capture


jmmans:
I guess we need to be very formal and someone review this so we can merge into main and start creating short-lived branches.

PR #232, opened 2025-11-05
disable menu printing, add `HELP` command instead

### LINKED ISSUE #227: recording menu interaction
I would really like it to be easier to record interactions within the menu. While printing the command options is helpful for interactivity, they get in the way when folks (I) want to copy-and-paste what I just did to show someone what happened.

*Solution*
Either hide the command listing behind some configuration of the menu _or_ print the command listing to `stderr` so `script` could write it to a different file?

tomeichlersmith:
This resolves #227 indirectly by switching the menu-printing to "request only" from "always on".

Not sure if other people want this, so I'm going to ask around before I merge it.

### DIFF
--- a/app/tool/tasks/align_phase_word.cxx
+++ b/app/tool/tasks/align_phase_word.cxx
@@ -296,7 +296,7 @@ void align_phase_word(Target* tgt) {
     // -------------- END SNAPSHOT BX SCAN ------------ //
 
   }  // -------- END WORD ALIGNMENT ------- //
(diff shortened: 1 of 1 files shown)


cbell1402:
I agree that the menu printing is a lot to look at every time we do something. Would there be a way to have the menu only print when we move from menu to menu? For example, I don't need the task sub-menu to print out over and over when I run a task but it would be helpful to see options when moving to a new sub-menu.

PR #233, opened 2025-11-05
have the register_maps library of LUTs compiled separately

tomeichlersmith:
this makes it so we only have to wait for that long compilation when one of the YAML files changes which /hopefully/ doesn't happen too frequently

### DIFF
--- a/register_maps/CMakeLists.txt
+++ b/register_maps/CMakeLists.txt
@@ -129,8 +129,6 @@ add_custom_command(
   COMMENT "Writing unifying header for parameter LUTs"
   VERBATIM
 )
-add_custom_target(register_maps_src
-  DEPENDS include/register_maps/register_maps.h register_maps.cxx)
 add_library(register_maps SHARED register_maps.cxx)
 target_include_directories(register_maps PUBLIC 
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>"


tomeichlersmith:
```
$ just clean configure
$ time just build
real    3m42.963s
user    0m0.189s
sys     0m0.165s
$ touch build/include/register_maps/register_maps.cxx
$ time just build
real    0m27.120s
user    0m0.092s
sys     0m0.072s
```

Changing the ROCs does not have as big of an effect, the total compilation time relative to the previous goes up by ~3s but this is within mesurement uncertainty. The fraction of time spent on the register maps remains about the same. However, this makes sense to me since the number of _entries_ in these register maps is vastly different.
```
tom@appa:~/code/ldmx/hgcroc/pflib/build/include/register_maps$ rg -c 'the_map\[' *.h
sipm_rocv3.h:213
si_roc.h:202
econd_test.h:18
si_rocv3b.h:206
sipm_rocv3b.h:218
econt.h:1331
econd.h:5303
```
I can believe that the compiler is able to have initializer lists approach the same speed as non-initializer lists for small lists and above some threshold it falls back onto a slow interpretation. In any case, I'm not going to worry about figuring that out. I still think this is a big improvemnt - both separating the compilation of the maps and reducing the time of that map compilation. Just going to try to organize the CMake a little bit and then merge it.

PR #237, opened 2025-11-10
Make opto abstract

jmmans:
This code change converts the opto interface into an abstract form, which can be implemented by multiple code designs. 

This development is not finalized completely, as the new code for the Bittware will be based from it, but this seems the right time to merge.


tomeichlersmith (approved):
This looks good, I'm going to apply the clang-format and then merge it.

PR #278, opened 2025-11-21
Adjust fastcontrol enables, econ align task

jmmans:
Creates a working ECON->lpGBT alignment task for ECON-D (single link), and adjusts the enables in Fast Control

PR #287, opened 2025-11-25
262 word alignment update

joshgreaves332:
Alignment task currently lets the user decide which roc board to manage.
Also allows user input for iecon number
3. Still to Do:
Update align_phase_word.cxx to determine whether we are connected by ZCU or bittware, and either search around bx or bx2560 respectively.
To achieve this, either leverage pftool::state flag, or accept user input for relevant BX number. (2560 vs 3534)
4 Make sure all eRx's are aligned and not to stop after the first one gets a match

### DIFF
--- a/src/pflib/ROC.cxx
+++ b/src/pflib/ROC.cxx
@@ -21,7 +21,6 @@ ROC::ROC(I2C& i2c, uint8_t roc_base_addr, const std::string& type_version)
 
 const std::string& ROC::type() const { return type_version_; }
 
-
 std::vector<uint8_t> ROC::readPage(int ipage, int len) {
   i2c_.set_bus_speed(1400);
 
@@ -57,8 +56,8 @@ uint8_t ROC::getValue(int ipage, int offset) {
     }
   }
 
-  PFEXCEPTION_RAISE("I2CFail",
-      "Failed to getValue after "+std::to_string(max_tries) +" attempts");
+  PFEXCEPTION_RAISE("I2CFail", "Failed to getValue after " +
+                                   std::to_string(max_tries) + " attempts");
 }
 
 std::vector<std::string> ROC::getDirectAccessParameters() {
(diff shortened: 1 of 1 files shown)


tomeichlersmith:
Many of these things are already on `main` from https://github.com/LDMX-Software/pflib/pull/278

https://github.com/LDMX-Software/pflib/blob/1231708d8bfd4578ba7f1079bee02fd349b69f3d/app/tool/tasks/align_phase_word.cxx#L46-L49

I think the biggest thing you should look at is getting the multi-link alignment to work. Not sure how to test that.

tomeichlersmith:
I just want to note here that #301 touched the same file that this PR is editing - focused on separating the bit alignment from the word alignment.

app/tool/tasks/align_phase_word.cxx
```diff
@@ -59,9 +59,13 @@ void align_phase_word(Target* tgt) {
 
   // Get channels from user
   std::string ch_str = pftool::readline(
-      "Enter channels (comma-separated), default is all channels: ",
+      "Enter channel(s) to check (comma-separated), default is all channels "
+      "0-7. Upon "
```
cmantill: I think that this should be 0-11

joshgreaves332:
Target objects are: HcalBackplane, EcalSMMTargetZCU, and EcalSMMTargetBW. 
Target.h now includes public virtual getRocErxMapping() and protected roc_to_erx_map_. 
HcalBackplane.h now includes public getRocErxMapping() override 
HcalBackplane.cxx now includes HcalBackplane::getRocErxMapping() to return protected member of class (mappings)
EcalSMM target cxx files include: getRocErxMapping() override  in the class definition due to absence of header file, and then a definition/implementation of getRocErxMapping() which returns EcalModule::getRocErxMapping(). These cxx files now also include ecal.h to facilitate this. 

Align phase word task now dynamically aligns the ROC channels according to the ROC which is selected, without prompting the user for input.

tvami (approved):

PR #293, opened 2025-12-02
ECON->lpGBT and ROC->ECON fixes

jmmans:
Also fixes in fast control, all ready for prime-time


tomeichlersmith:
All of the commits on this branch (except the last one) were on `main`, so I've rebased the branch onto main only keeping the last commit.

PR #312, opened 2025-12-10
311 make it easier to set bus speed and so do for bias

### LINKED ISSUE #311: Make it easier to set bus speed and so do for bias
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

jmmans:
Makes the lpGBT bus speed setting simpler and applies 100 kHz to all the Bias interactions.  Probably should change code to restore whatever speed was there before for the non-bias interactions (e.g. ROC)

### DIFF
--- a/include/pflib/lpGBT.h
+++ b/include/pflib/lpGBT.h
@@ -208,26 +208,36 @@ class lpGBT {
 
   uint32_t read_efuse(uint16_t addr);
 
-  /** Setup an I2C bus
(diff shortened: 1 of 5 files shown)


jmmans:
I think this can still go in, but the "get speed" needs to be better-handled.

tomeichlersmith (approved):

Issue #315, opened 2025-12-11
GPIOs are not set for lpGBT inside pftool menu


cmantill:
e.g. ECON->HARDRESET results in:
```
 > hardreset
(2025-12-10 21:13:20) [menu] error: [GPIOError] : Unknown GPO bit 'ECON_HRST'
```

PR #316, opened 2025-12-11
Enforce running standard configs so that the GPIOs are set

tvami:
Resolves https://github.com/LDMX-Software/pflib/issues/315

Even if the PUSM is correct, we still need to run the STANDARD configs, otherwise the GPIOs are not set.


tomeichlersmith (approved):
makes sense to me, thank you for the investigation

PR #336, opened 2025-12-12
Fix  HT_EBO, data underflow for no data

tvami:
There is two real things (+ internal documentation) in this PR:
 
I down the rabbit hole about what these packets mean, but Fig 33 is great!!
Screenshot 2025-12-12 at 00 00 53

So based on that I found that 
* `ht_ebo` is on 4 bits, not 5 so we need to do
```
uint32_t ht_ebo = ((data[0] >> 8) & mask<5>);
```
to
```
uint32_t ht_ebo = ((data[0] >> 8) & mask<4>);
```

And then I copy pasted a bunch of text from the document so the next person doesnt have to figure all this out.

EDIT: Some testing gave me 
```
warn: Incomplete event packet, stored payload length 1 is larger than the packet length 18446744073709551614
```
which is uint(0-2), so the last commit now adds +2 on the other side and then the printouts let the users do "math"

### DIFF
--- a/src/pflib/packing/ECONDEventPacket.cxx
+++ b/src/pflib/packing/ECONDEventPacket.cxx
@@ -283,7 +283,7 @@ void ECONDEventPacket::from(std::span<uint32_t> data) {
   uint32_t hamming = (data[0] & mask<6>);
   pflib_log(trace) << "    P=" << passthrough << " E=" << expected;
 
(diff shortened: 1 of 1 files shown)


tvami:
@tomeichlersmith I removed the change for the `MASK_IO_SIZE_NEXT` I didnt realize that's a firmware thing. Would taking to Jeremy help that? Anyway, let's have it in a separate PR

PR #337, opened 2025-12-12
Cross talk - Multi Channel Scan

andpet1324:
New task `multi_channel_scan` allows one to scan and analyze all channels at the same time, for an increasing number of pulsed channels per iteration. New function `adc_all_channels` in `ana/charge/parameter_timescan.py` plots the ADC vs channels and ADC vs time, including some RMS calculations for the output data. In the task function, there's the option of doing external charge injections with the LED L1A trigger, otherwise you can chose from preCC and highrange injections.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,10 @@ name: test
 
 on:
   push:
+  workflow_run:
+    workflows: ["clang-format"]
+    types:
+      - completed
 
 jobs:
   test:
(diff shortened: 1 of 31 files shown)


tomeichlersmith (commented):
Looks good, I just have two comments to help make the code more widely usable.

If the one about the decoding is too complicated, I'm also willing to just merge this and then I'll update it on my refactoring/tidying pass.

PR #345, opened 2025-12-14
futsing around with charge_timein while trying to debug it

tomeichlersmith:
- increase number of tries in the ROC::getValue and add a sleep on fails
- write a maker that returns a `std::unique_ptr` so all the run logic isn't copied
- call `setup_run` before each change to the fc calib
- move from a lambda function to a free function in `daq`

I don't know if any of this is necessary to get charge injection working. I actually strongly suspect its not, but it is what I did to the code while working on it.

### DIFF
--- a/app/tool/daq.cxx
+++ b/app/tool/daq.cxx
@@ -666,8 +666,9 @@ static void daq_debug_trigger_timein(Target* tgt) {
       "Want to try another set of timing parameters?", false));
 }
 
-template<class EventPacket>
-std::unique_ptr<DecodeAndWriteToCSV<EventPacket>> make_all_channel_writer(const std::string& file_name) {
+template <class EventPacket>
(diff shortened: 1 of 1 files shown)


app/tool/daq.cxx
```diff
@@ -666,6 +666,65 @@ static void daq_debug_trigger_timein(Target* tgt) {
...
+    case Target::DaqFormat::SIMPLEROC:
+      return make_all_channel_writer<pflib::packing::SingleROCEventPacket>(
+          file_name);
+      break;
+    default:
+      PFEXCEPTION_RAISE("NoDecode",
+                        "Unable to decode format currently configured.");
+      break;
+  }
+}
```
tomeichlersmith: This is the only thing I really want to keep. The other stuff is probably (hopefully) not necessary.

PR #351, opened 2025-12-15
305 lpgbt temperature

cbell1402:
Solves 305 by estimating the internal temperature of the lpGBT board. 

1. Reads the lpGBT CHIPID from the efuses
2. Checks if a `calibration_results.yaml` file already exists
3. Checks if the CHIPID is in that file
4. If not or if the file doesn't exist, opens the `lpgbt_calibration.csv` file (downloads the csv file if needed) and finds the constants
5. Finds temperature of the lpGBT internal sensor based on some formulas using the calibrations constants and ADC measurements
6. Prints the temperature for the user

Print out for if the correct constants already exist in the yaml file:
```
 > GET_LPGBT_TEMPS
 --- Current lpGBT CHIPID: CC2BDEC9 ---
  > Found existing calibration file 'calibration_results.yaml'...
  > Inferred Junction Temperature (TJ_USER) from vref_tune: 45.366795 degrees C
  > Updated 'calibration_results.yaml' with TJ_USER
  > VREFTUNE REG: 103
  > Calibrated Voltage: 0.5049 V
  > ----------------------------------------------------
  > lpGBT Internal Temperature: 25.94 degrees C
  > ----------------------------------------------------
```
I followed Gabi Hamilton's slides very closely for this
https://github.com/user-attachments/files/23972425/CassettesTemps_CWM_Oct15.pdf

### DIFF
--- a/app/tool/tasks/get_lpgbt_temps.cxx
+++ b/app/tool/tasks/get_lpgbt_temps.cxx
@@ -134,7 +134,7 @@ void get_lpgbt_temps(Target* tgt) {
     bool verify = pftool::readline_bool(
         "Using precise constants may require downloading and unzipping a large "
         "CSV file (330 MB) - Continue?",
-        false);
+        true);
     if (!verify) {
       printf("  > Continuing with average constants...\n");
       lpgbt_card.read_internal_temp_avg();
--- a/src/pflib/lpGBT.cxx
+++ b/src/pflib/lpGBT.cxx
@@ -386,15 +386,15 @@ int lpGBT::find_vref_tune(double vref_slope, double vref_offset,
 
 void lpGBT::read_internal_temp_avg() {
   double adc_avg = adc_average();
-  printf("  > ADC Average: %f\n", adc_avg);
   double tj_user = adc_avg * AVG_TEMPERATURE_UNCALVREF_SLOPE +
                    AVG_TEMPERATURE_UNCALVREF_OFFSET;
-  printf("  > Esitimated junction temperature (TJ_USER): %f degrees C\n",
-         tj_user);
-
   double vadc_v =
       adc_avg * (AVG_ADC_X2_SLOPE + tj_user * AVG_ADC_X2_SLOPE_TEMP) +
       AVG_ADC_X2_OFFSET + tj_user * AVG_ADC_X2_OFFSET_TEMP;
+
+  printf("  > ADC Average: %f\n", adc_avg);
+  printf("  > Esitimated junction temperature (TJ_USER): %f degrees C\n",
+         tj_user);
   printf("  > Calibrated Voltage: %.4f V\n", vadc_v);
 
   double temperature_c =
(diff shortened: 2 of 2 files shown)


cbell1402:
Yes, they are like serial numbers. To my knowledge, the DAQ and TRG lpGBTs are the same but would have different chip IDs and also likely different constants. 

There is some more information here
https://lpgbt.web.cern.ch/lpgbt/v1/analog.html#calibration-data-format-and-distribution

Also, it looks like there are "typical" constants listed here
https://lpgbt.web.cern.ch/lpgbt/v1/electricalCharacteristics.html#temperature-sensor-characteristics

So, perhaps I could do some testing to see if we get reasonable temperature readings with these default constants. Perhaps reasonable could be defined as +- a few degrees C. The only issue is, I would need to know the actual temperature from someone physically present at the mezzanines.

tomeichlersmith:
Unfortunately, I do not have access to that area of `cern.ch` (probably because I'm not in the proper "HGCAL Electronics" egroup or something).

Anyways, I do have physical access to the HcalBackplane and its lpGBTs here at UMN so I'd just need to find a temp sensor that I can use to verify that the "typical" constants are good enough.

I still think your detailed code is good especially in the long term where we would have a specific set of well-tested lpGBTs. This would then give us a database of calibration constants to either look-up live or cache in a local CSV/DB/YAML file. Just trying to think about how to structure it so it can fail to do the precise calculation quickly and quietly.

tomeichlersmith:
Yes, have the precise calibrations that require download "opt-in"

cbell1402:
I moved the temperature calculations to lpGBT.cxx and lpGBT.h. The task now asks the user if they would like to use averaged constants to find the temperature. The averaged constants come from averaging together every row of data in the calibration database for all lpGBT cards. If the user says yes, then the hard-coded average constants are used. If they say no, then it asks them if they are sure before either:

- Looking for an existing YAML file with constants for the chipid in question
- If the YAML doesn't exists or doesn't have the correct chipid, it pulls constants from the .csv file
- If the .csv files doesn't exist, it downloads it

The code then continues to calculate internal temperature. 

I found that the calculated temperature using the averaged constants is usually within 5 degrees C of the precise constants. 

Below is what running the task looks like.

**Averaged constants**
```
 > GET_LPGBT_TEMPS
 --- Current lpGBT CHIPID: CC2BDEC9 ---
Use average constants? (Y/N)  [Y]
  > ADC Average: 513.050000
  > Esitimated junction temperature (TJ_USER): 23.379317 degrees C
  > Calibrated Voltage: 0.5013 V
  > ----------------------------------------------------
  > lpGBT Internal Temperature: 21.78 degrees C
  > ----------------------------------------------------
```

**Precise constants**
```
 > GET_LPGBT_TEMPS
 --- Current lpGBT CHIPID: CC2BDEC9 ---
Use average constants? (Y/N)  [Y] n
Using precise constants may require downloading and unzipping a large CSV file (330 MB) - Continue? (Y/N)  [Y]
  > Found calibration constants...
  > Optimal VREFTUNE: 103;  Writing to REG_VREFTUNE...
  > ADC Average: 512.900000
  > Esitimated junction temperature (TJ_USER): 45.366795 degrees C
  > Calibrated Voltage: 0.5038 V
  > ----------------------------------------------------
  > lpGBT Internal Temperature: 25.49 degrees C
  > ----------------------------------------------------

```

tomeichlersmith (approved):
alright, just tested at UMN and it works there too, so I'm going to merge this

I may end up wanting to move this code around, but it is at least available now. Thank you!

PR #363, opened 2026-01-28
Mean stdev andreas

andpet1324:
Added two utility functions: one calculating the mean of a list of samples, and one calculating the stdev. For use in `inv_vref` parameter algorithm.

### DIFF
--- a/src/pflib/utility/stdev.cxx
+++ b/src/pflib/utility/stdev.cxx
@@ -1,8 +1,8 @@
 #include "pflib/utility/stdev.h"
 
-#include <cmath>    // std::sqrt, std::pow, std::abs
-#include <numeric>  // std::accumulate
+#include <cmath>  // std::sqrt, std::pow, std::abs
 #include <limits>
+#include <numeric>  // std::accumulate
 
 #include "pflib/utility/mean.h"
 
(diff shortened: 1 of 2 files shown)


include/pflib/utility/mean.h
```diff
@@ -0,0 +1,16 @@
...
+namespace pflib::utility {
+
+/**
+ * find the mean of the input vector of samples
+ *
+ * @param[in] samples list of samples to find mean of
+ * @return mean of the samples
+ */
+double mean(std::vector<int> samples);
+double mean(std::vector<double> samples);
```
tomeichlersmith: Please add documentation to both implementations.

src/pflib/utility/stdev.cxx
```diff
@@ -0,0 +1,36 @@
...
+  if (n == 0) {
+    throw "Trying to take the stdev of an empty vector";
+  }
+  double mean = pflib::utility::mean(samples);
+  std::vector<double> d2;
+  for (int i = 0; i < n; i++) {
+    d2.push_back(std::pow(std::abs(samples[i] - mean), 2));
+  }
+  return std::sqrt(std::accumulate(d2.begin(), d2.end(), 0.0) / n);
+}
```
tomeichlersmith: same comments as mean (try `template` instead of just copying the code, nan instead of exception, const reference), but I also don't know why you are creating another vector `d2`. This is an extra vector that we don't use later and so its just an extra loop with extra memory overhead.

PR #368, opened 2026-02-15
fixed indice bug

andpet1324:
Bug that caused incorrect TOA_VREF on link 1. An array was called outside of its assigned size.


tomeichlersmith (approved):
Yep, that'll do it, good catch

PR #391, opened 2026-04-03
Default iroc to first valid ROC index from boardmask

OwenCole04:
iroc defaults to 0 which is invalid when ROC 0 isn't included. Now sets iroc to first ROC specified by the boardmask upon starting pftool.


tomeichlersmith (approved):
great 😎 i've been meaning to do this too, thank you

PR #395, opened 2026-04-13
Pedestal averages

blazeiro:
I made a short script for evaluating the pedestals of all the channels in a given board, inspired by the violinplot script. The reason for this addition is that the violinplot algorithm does not show the _average_ pedestal ADC code for each channel. Furthermore, the errors in the violin-points are difficult to compare by eye. For comparison, below I attached the violinplot and averagesplot script outputs for link 0, respectively:

pedestal-0-B24-violinplot

pedestal_avg_link0

The pedestals-0 and pedestals-1 in the second plot are pedestals before and after alignment, the names of which can be adjusted.


ana/pedestal/averagesplot.py
```diff
@@ -0,0 +1,116 @@
...
+        ymax.append([max(sample_avg_adc[n][0].ravel()),max(sample_avg_adc[n][1].ravel())])
+        ax1.errorbar(channels, sample_avg_adc[n][0].ravel(), yerr=sample_error_adc[n][0].ravel(), label = f'Pedestals {n}, link 0', fmt='o')
+        ax2.plot(channels, sample_error_adc[n][0].ravel(), label = f'Pedestals {n} stdev, link 0')
+
+    ymin = np.array(ymin)
+    ymax = np.array(ymax)
+    ax1.set_ylim([min(ymin[:,0]-10), max(ymax[:,0])+10])
+
+    ax1.set(ylabel='ADC [a.u.]', xlabel='Channels')
+    ax2.set(ylabel='ADC [a.u.]', xlabel='Channels')
```
tomeichlersmith: Please add more detail to the ylabel so that the plot does not require as much explanation. In this case, you could say "Mean ADC" for `ax1` and `Std Dev ADC` or `ADC RMS` for `ax2`.

(Also do this for ax3/ax4 below)

blazeiro: Ok! Here is the updated output:

pedestal_avg_link0

tomeichlersmith (approved):

PR #396, opened 2026-04-13
393 noinv alignment

### LINKED ISSUE #393: Better selection criteria for tasks/inv_vref_scan_lund
**Is your feature request related to a problem? Please describe.**
In its current state, the algorithm to align `INV_VREF` makes some assumptions about the behavior of the chip, which manifests in the selection criteria. We want to improve the selections by studying how different boards behave. I would also like to see some fail-safes to catch when a board is "misbehaving". 

The current assumptions are based on the figures below:

Image

This figure shows the parameter space of `INV_VREF` and `NOINV_VREF`. There are clear distinct regions, and we want to select a value somewhere in the linear region between `INV_VREF` of about 200 - 600. To replicate this figure, run `TASKS/VREF_2D_SCAN` and analyze the data with `ana/heatmap.py`. From this figure we can't make any qualitative selections. In the next figure I've plotted the gradient of each point in `INV_VREF`.

Image

The gradient is:

$$
 y(x) = \frac{ADC(x) - ADC(x-1)}{inv_{vref}(x) - inv_{vref}(x-1)}.
$$

From this figure we can select the linear region based on that the point's gradient should have
1. $` abs(y(i) - median(y)) < 0.7 \cdot abs(median(y)) `$
3. standard deviation > 0

These selections are based on the information from board 21 at Lund, and are not very general. These selections have worked for further testing of 9 more boards, however they do not work for board 20 where the region below ~200 have a lot of noise, shown in the following figure

Image

For this board, it might be enough to add an upper limit on the noise. The whole procedure to align `INV_VREF` is heavily inspired by [CMS's alignment](https://github.com/Mekhpar/Tileboard_processing/blob/main/vref2D_scan_analysis.py), however I find that their selections do not align with our boards at times. They also do some more complex selections that could be useful.

**Describe the solution you'd like**
Further study of the `INV_VREF` and `NOINV_VREF` parameter space on different boards. Testing more complex selections from CMS's alignment procedure. Fail-safes to catch when a board isn't behaving as assumed, and is therefore failing the selections.

andpet1324:
Adds noinv_vref alignment to existing inv_vref alignment. Caught by bad fits and boundary cases.

### DIFF
--- a/app/tool/algorithm/inv_vref_lund.cxx
+++ b/app/tool/algorithm/inv_vref_lund.cxx
@@ -161,6 +161,7 @@ std::map<std::string, std::map<std::string, uint64_t>> inv_vref_lund(
     Target* tgt, ROC& roc) {
   static auto the_log_{::pflib::logging::get("inv_vref_scan")};
   int nevents = pftool::readline_int("Number of events per point: ", 100);
+  int noinv_vref_step = pftool::readline_int("Stepsize for noinv_vref: ", 20);
   // TODO 348
 
   std::array<int, 2> inv_vref_tgt;
@@ -195,7 +196,7 @@ std::map<std::string, std::map<std::string, uint64_t>> inv_vref_lund(
     inv_vref_tgt[1] = fitter_l1.fit(target_adc);
 
     if ((inv_vref_tgt[0] == -1) || (inv_vref_tgt[1] == -1)) {
-      noinv_vref -= 20;
+      noinv_vref -= noinv_vref_step;
       pflib_log(info) << "Bad slope and/or intercept. "
                       << "Setting noinv_vref to " << noinv_vref;
       continue;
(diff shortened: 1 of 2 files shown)


app/tool/algorithm/inv_vref_lund.cxx
```diff
@@ -165,13 +149,61 @@ std::map<std::string, std::map<std::string, uint64_t>> inv_vref_lund(
...
+    // sort data and fit
+    DataFitter fitter_l0;
+    DataFitter fitter_l1;
+    fitter_l0.sort_and_append(inv_vrefs, pedestals_l0, stds_l0, step);
+    fitter_l1.sort_and_append(inv_vrefs, pedestals_l1, stds_l1, step);
+    try {
+      inv_vref_tgt[0] = fitter_l0.fit(target_adc);
+      inv_vref_tgt[1] = fitter_l1.fit(target_adc);
+      break;
+    } catch (...) {
```
tomeichlersmith: Please add a specific type to the exception you are catching. As written, we could (for example) have a bug in `DataFitter` that is causing an `std::out_of_range` exception but we are treating that like a "bad fit" exception. I would like `DataFitter` to be updated to be able to inform the user when the fit fails (what exception is being thrown when a fit "fails"?)

andpet1324: I realized that I was doing this wrong. I added cases instead of finding a specific exception.

tomeichlersmith (approved):
Cool, this makes sense to me, thank you

PR #399, opened 2026-04-20
unshuffle links in level pedestals

tomeichlersmith:
Starting with level pedestals since its the first ROC calibration that we need to do anyways. It also is the simplest (in some sense) since we don't need charge injection working.

### DIFF
--- a/app/tool/algorithm/level_pedestals.cxx
+++ b/app/tool/algorithm/level_pedestals.cxx
@@ -144,6 +144,10 @@ level_pedestals(Target* tgt) {
       lowend[i_roc] =
           get_adc_medians(i_roc, tgt->getRocErxMapping(), buffer.get_buffer());
     }
+
+    for (int i_roc : tgt->roc_ids()) {
+      tgt->roc(i_roc).setRegisters(prior_registers.at(i_roc));
+    }
   }
 
   pflib_log(info) << "sample collections done, deducing settings";
--- a/app/tool/algorithm/level_pedestals.cxx
+++ b/app/tool/algorithm/level_pedestals.cxx
@@ -41,14 +41,18 @@ static int get_adc(const pflib::packing::MultiSampleECONDEventPacket& p,
  * TOT/TOA and the sample Tp/Tc flags are ignored.
  */
 static std::array<int, 72> get_adc_medians(
-    const std::vector<pflib::packing::MultiSampleECONDEventPacket>& data) {
+    int i_roc,
+    const pflib::packing::SingleECONDRocErxMapping& mapping,
+    const std::vector<pflib::packing::MultiSampleECONDEventPacket>& data
+  ) {
   std::array<int, 72> medians;
   /// reserve a vector of the appropriate size to avoid repeating allocation
   /// time for all 72 channels
   std::vector<int> adcs(data.size());
   for (int ch{0}; ch < 72; ch++) {
     for (std::size_t i{0}; i < adcs.size(); i++) {
-      adcs[i] = get_adc(data[i], ch);
+      auto [i_erx, i_ch] = mapping.toErxChannel(i_roc, ch);
+      adcs[i] = data[i].soi().channel(i_erx, i_ch).adc();
     }
     medians[ch] = pflib::utility::median(adcs);
   }
--- a/app/tool/algorithm/level_pedestals.cxx
+++ b/app/tool/algorithm/level_pedestals.cxx
@@ -65,11 +65,7 @@ level_pedestals(Target* tgt) {
       parameters[ch_str]["DACB"] = 0;
       parameters[ch_str]["TRIM_INV"] = 0;
     }
-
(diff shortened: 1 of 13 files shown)


tomeichlersmith:
I am having trouble getting a reliable connection to the EcalSMM here at UMN (hopefully just due to the old ROCv2a we are using, soon to be replaced), so I tested this multi-ROC refactor on a single-ROC HcalBackplane. This at least confirms that the unshuffling of the ECON links works as intended (at least, I'm assuming that if we were looking at data for Ch 2 from Link 0 to determine the settings for Ch 2 on Link 1 the resulting parameters would be a lot less level).

Left is the default pedestals and right are the ones after running `tasks.level_pedestals` on this branch.

hcal-backplane-default-pedestals

hcal-backplane-level-pedestals

:warning: *The plotting is probably mixing up the Link 0/1 shuffle that is imposed by the ECON-D on the HcalBackplane.* I think we can make due with this right now, but I would like to look at switching to the ROC/Channel indexing for output CSVs compared to the (current) "i_link/channel" indexing which has the ECON-D shuffle imbued within it.

app/tool/tasks/level_pedestals.cxx
```diff
@@ -7,15 +7,19 @@
...
     out << YAML::Value << YAML::BeginMap;
-    for (const auto& param : page.second) {
-      out << YAML::Key << param.first << YAML::Value << param.second;
+    for (const auto& page : roc.second) {
+      out << YAML::Key << page.first;
+      out << YAML::Value << YAML::BeginMap;
+      for (const auto& param : page.second) {
+        out << YAML::Key << param.first << YAML::Value << param.second;
+      }
+      out << YAML::EndMap;
```
tomeichlersmith, line 20: The YAML constructed by this code is not currently parse-able by pflib. It is closer to a "subsystem config" where mutliple different ROC's configuration parameters are stored the same file under their indices as keys.

Does anybody actually dump this stuff to a file? I just view it and then if it looks good, apply it.

app/tool/algorithm/level_pedestals.cxx
```diff
@@ -66,127 +49,173 @@ std::map<std::string, std::map<std::string, uint64_t>> level_pedestals(
...
-    lowend = get_adc_medians(buffer.get_buffer());
+
+    for (int i_roc : tgt->roc_ids()) {
+      lowend[i_roc] =
+          get_adc_medians(i_roc, tgt->getRocErxMapping(), buffer.get_buffer());
+    }
+
+    for (int i_roc : tgt->roc_ids()) {
+      tgt->roc(i_roc).setRegisters(prior_registers.at(i_roc));
+    }
```
tomeichlersmith: This is yucky to me since I've already forgotten it and confused myself for a few hours. I'm thinking about writing another class like `TempParameters` but which can be used for a set of ROCs.

app/tool/algorithm/level_pedestals.cxx
```diff
@@ -66,127 +49,173 @@ std::map<std::string, std::map<std::string, uint64_t>> level_pedestals(
...
+    std::map<std::string, std::map<std::string, uint64_t>> parameters;
     for (int ch{0}; ch < 72; ch++) {
-      std::string page{"CH_" + std::to_string(ch)};
-      test_handle_builder.add(page, "SIGN_DAC", 0);
-      test_handle_builder.add(page, "DACB", 0);
-      test_handle_builder.add(page, "TRIM_INV", 0);
+      std::string ch_str{"CH_" + std::to_string(ch)};
+      parameters[ch_str]["SIGN_DAC"] = 0;
+      parameters[ch_str]["DACB"] = 0;
+      parameters[ch_str]["TRIM_INV"] = 0;
```
tomeichlersmith, line 66: While I didn't get to the point of actually running the leveling algorithm over successful pedestal runs, I did notice a difference between the Silicon HGCROC chips (for ECal) and the SiPM chips (for Hcal).

The `si_rocv3b` parameters do **not** contain `DACB` and `SIGN_DAC` in each channel. It does have `TRIM_INV` though which makes me think that we should have some part of the pedestal leveling happening only with using `TRIM_INV` (using `32` as the middle value?) and then `SIGN_DAC`+`DACB` can be used for SiPM HGCROCs to further tune the pedestals (since we expect the SiPM chips to have more noise when connected to their CMBs etc...?)

tomeichlersmith: Re-reading the HGCROC tuning presentation from our CMS colleagues convinces me that we _should_ be using `TRIM_INV: 32` as the middle value even in the SiPM HGCROC.

tomeichlersmith: Including this patch in a later pedestal leveling refactor: https://github.com/LDMX-Software/pflib/issues/407

tomeichlersmith (commented):
While this hasn't been tested on multiple ROCs, the fact that it is functional with the single ROC on an HcalBackplane is giving me more confidence that it is good enough to prime time.

PR #408, opened 2026-05-01
Pedestal-averages (#395) plot changes

blazeiro:
I made some changes and additions to the pedestal-averages plotting script, whose previous output was as follows: 

pedestals_old

The plot used standard deviation for errors of the mean-pedestal values, which have now been changed to root-mean-square deviation (RMS). Furthermore, to make it more clear that the error values are not continuous in the RMS-graph, I changed the opacity of the line and added markers for each point. I also changed the grid of the RMS-graph to make it easier to read the x-axis, rather than the y-axis. Lastly, I added a check that informs the script that the first data file corresponds to pre-alignment pedestals, while the second to post-alignment data (though more than two files can be processed and the setting is entirely optional). These changes result in the following figure (per link):

pedestal_avg_link0

I also added a small function to calculate the RMS of the mean-pedestal distribution within each link, saving the results to a csv file. Since the pedestals after-alignment should settle at, ideally, the same value on each channel per link, their distribution can be treated as one of samples from the same measurement. The RMS of this distribution can be used as a quick diagnostics-check in other scripts or when testing multiple boards, in lieu of a great number of plots.


ana/pedestal/averagesplot.py
```diff
@@ -22,6 +24,13 @@
 sample_avg_adc = []
 sample_error_adc = []
 
+def root_mean_square(adc):
+    mean_adc = np.mean(adc)
+    sum = 0
+    for sample in adc:
+        sum += (sample-mean_adc)**2
+    return np.sqrt(sum/len(adc))
```
tomeichlersmith: Just to be clear about the vocabulary, the RMS (when subtracting the mean from each sample) _is_ the standard deviation. I don't have a problem with leaving the "RMS" nomenclature in since that is more familiar us in the context of electronics noise, but we should use the `np.std` function instead of this one since the `np.std` function is faster and more well tested.

tomeichlersmith (changes requested):
Just one requested change which I can do for you if you're confused. Just let me know.

ana/pedestal/averagesplot.py
blazeiro: You're completely right, I didn't realise RMS deviation is the same as StdDev for some reason. It should be good now!

tomeichlersmith (approved):

PR #409, opened 2026-05-01
Time-scan mean-ADC-peak evaluation

blazeiro:
In the same vein as the average-pedestal plotting script (#408), I thought a script comparing the mean-ADC-peaks of channels per link and CALIB would make for a good diagnostics feature. In short, the ideal behavior of the ADC should result in ADC-peaks coalescing around the same value within each link (same charge-injection = same response). This script takes in data from different CALIBs and plots the results, mean-ADC-peaks, on the same figure for comparison. An additional figure is made to track the RMS of each mean-peak.

The script takes in (ideally at least 72, but can accommodate less) time-scan data files corresponding to the same channel and CALIB value, as well as time-scan mode (highrange, lowrange, preCC). More CALIB-sets can be added for comparison, but this is not strictly necessary. After this, the script computes the mean of ADC-peak samples on each channel, along with their RMS. Four plots in total are made, with the following two corresponding to one link:

time-scan-evaluation-highrange

time-scan-evaluation-rms-highrange

The script can serve as an HGCROC-wide ADC evaluation tool, showing whether any channel is behaving strangely, whether at some specific CALIB-value or across multiple. Additionally, an RMS calculation of the peak-distributions on each link per CALIB is available, with the option to save the results to a csv.

 As a small note, depending on the gain, the ADC will likely saturate after some given CALIB during highrange scan. The peaks of those scans are often constant (1023) within each sample distribution, leading to null-RMS. This can be seen in the figures above, where CALIB 512 is clearly saturated on all channels except 19. This is accounted for in the script by skipping any set with a fully-null RMS distribution.


tomeichlersmith (approved):
Looks good and makes sense to me, thank you

PR #410, opened 2026-05-01
Actual changes to split out the TRG link

### LINKED ISSUE #406: ECON-T Alignment and Configuration
Since we were able to observe data on the trigger links from an HcalBackplane, Jeremy has been working on updating the firmware to be able to observe this data and write a rudimentary self-trigger, but this needs to go together with two features here in pflib. This is me expanding on notes shared by @cmantill 

## ECON-T Alignment
The ECON-T alignment is slightly different than the ECON-D alignment.

### Serializer (or ECON-lpGBT) 
Bit Alignment
- should activate PRBS7 on the output with `FormatterBuffer.Global.etx_pattern = 1`, the setting below is only for the ECON-D
- update to do bit alignment on all active eTx from ECON-D or ECON-T, our ECON-Ds only have one eTx but the ECON-Ts will have 3 (for HcalBackplane) and potentially more for Ecal
  - probably extend `TRIG` to hold the mapping of ECON-T eTx to lpGBT input elinks

https://github.com/LDMX-Software/pflib/blob/30f792c6a25fb37e36322bbc18a6e3ea8a5bdf92/app/tool/tasks/setup/align_econ_lpgbt.cxx#L55

Word alignment between ECON-T and lpGBT
- same in principle although again we need to do it on all active eTx

## ECON-T Configuration
The ECON-T has many different output modes and, after discussion with Jeremy and Cristina, we were able to deduce a good guess at what the output mode settings for the ECON-T should be for the Ecal and Hcal.

### Hcal
- 3 eTx: `FormatterBuffer.Global.active_etx = 0b111`
- 8 eRx: `erx.NN_enable` (exact `NN` set depend on indexing and which ROCs/channels are in use)
- might want to use the multiplexer later to map e.g. ECON-T-0 eRx0 to ROC0 TRG instead of ROC1-TRG3 (the current default), this is interwoven with cabling since our goal is to have both ends of a quad bar summed into super trigger cells (STCs) by the ECON-Ts.
- have 5E+4M encoding using all 9 bits: `AlgoDroplsb.Global.drop_lsb = 1`
- select STC algorithm: `ConfigAlgo.Global.alg_type = 1`
- `FormatterBuffer.global.stc_type = 0` ???

### Ecal
- 12 eRx: `erx.NN_enable = 1` for all `NN`:6 ROCs each with 2 halves -> all 12 eRx in use (like the ECON-D)
- 4? eTx: `FormatterBuffer.Global.active_etx = 0xf`, not confident in this guess
- same comments about multiplexer since not sure how the ROCs are indexed in the ECON-T
- determine encoding for output results
- we will want the "Best Choice" algorithm (which returns a sum over all input trigger cells (TCs) which is a whole module in our design and the N highest energy TCs where N depends on the number of eTx/eRx). `ConfigAlgo.Global.alg_type = 3`
- since we have HD modules, we'll want to select the high density version of the Best Choice algo: `ConfigAlgo.Global.alg_high_density = 1`

tomeichlersmith:
First step of viewing TRG link is making sure we are appropriately viewing its status when doing resets to try to get it working...

### DIFF
--- a/src/pflib/zcu/zcu_optolink.cxx
+++ b/src/pflib/zcu/zcu_optolink.cxx
@@ -124,7 +124,7 @@ std::map<std::string, uint32_t> ZCUOptoLink::opto_status() {
   retval["CDR_LOCK"] = transright_.read(7) & 0xFFF;
 
   val = coder_.read(2);
-  prefix = "LINK" + std::to_string(ilink_);
+  std::string prefix = "LINK" + std::to_string(ilink_);
   retval[prefix + " READY"] = (val >> (0 * 2 + (ilink_ % 2))) & 0x1;
   retval[prefix + " NOT_IN_RESET"] = (val >> (1 * 2 + (ilink_ % 2))) & 0x1;
   retval[prefix " LINK_ERRORS"] = coder_.read(4 + (ilink_ % 2)) & 0xFFFFFF;
--- a/src/pflib/zcu/zcu_optolink.cxx
+++ b/src/pflib/zcu/zcu_optolink.cxx
@@ -124,9 +124,9 @@ std::map<std::string, uint32_t> ZCUOptoLink::opto_status() {
   retval["CDR_LOCK"] = transright_.read(7) & 0xFFF;
 
   val = coder_.read(2);
-  prefix = "LINK"+std::to_string(ilink_);
-  retval[prefix+" READY"] = (val >> (0 * 2 + (ilink_ % 2))) & 0x1;
-  retval[prefix+" NOT_IN_RESET"] = (val >> (1 * 2 + (ilink_ % 2))) & 0x1;
+  prefix = "LINK" + std::to_string(ilink_);
+  retval[prefix + " READY"] = (val >> (0 * 2 + (ilink_ % 2))) & 0x1;
+  retval[prefix + " NOT_IN_RESET"] = (val >> (1 * 2 + (ilink_ % 2))) & 0x1;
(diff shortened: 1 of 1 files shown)


src/pflib/zcu/zcu_optolink.cxx
```diff
@@ -62,13 +62,15 @@ void ZCUOptoLink::reset_link() {
    *
    * Unsure if this should depend on ilink. Currently, it does not.
```
tomeichlersmith: update comment before merging

src/pflib/zcu/zcu_optolink.cxx
```diff
@@ -122,9 +124,9 @@ std::map<std::string, uint32_t> ZCUOptoLink::opto_status() {
   retval["CDR_LOCK"] = transright_.read(7) & 0xFFF;
 
   val = coder_.read(2);
-  retval["READY"] = (val >> 0) & 0x1;
-  retval["NOT_IN_RESET"] = (val >> 1) & 0x1;
-  retval["LINK_ERRORS"] = coder_.read(4) & 0xFFFFFF;
+  retval["READY"] = (val >> (0 * 2 + (ilink_ % 2))) & 0x1;
+  retval["NOT_IN_RESET"] = (val >> (1 * 2 + (ilink_ % 2))) & 0x1;
+  retval["LINK_ERRORS"] = coder_.read(4 + (ilink_ % 2)) & 0xFFFFFF;
```
tomeichlersmith: maybe add `ilink` to name of value to emphasize to user that this is per-link (while the other stuff is per-quad)?

PR #414, opened 2026-05-11
Outlier-scan general changes and improvements

blazeiro:
To make the scan a bit more useful, I added "bulk-evaluation" functionality, i.e. the ability to analyze many data files at once. Specifically, I added the option to analyze and plot the total outlier number and frequency of samples containing any outliers, across multiple channels and CALIB values. The two plots are histograms, and look something like this:

outlier_evaluation_sum

outlier_evaluation_freq

The plots can include other CALIB values at the same time.

Other than that, I made some general improvements to the figures and changed the thresholds for the default outlier identification algorithm. I also added another outlier identification method based on using standard deviation. I will write up a wiki entry for all of this soon!


tomeichlersmith (approved):
looks good to me, thank you

PR #416, opened 2026-05-14
ECON-T/ROC Alignment

### LINKED ISSUE #406: ECON-T Alignment and Configuration

tomeichlersmith:
This updates the ECON/ROC alignment task to support ECON-T
## To Do
- [x] find a place to store mapping of ROC TRG to ECON-T eRx from a specific hardware
- [x] define EcalSMM mapping using schematic as reference

### DIFF
--- a/app/tool/main.cxx
+++ b/app/tool/main.cxx
@@ -369,7 +369,7 @@ int main(int argc, char* argv[]) {
       tgt.reset(pflib::makeTargetHcalBackplaneBittware(ilink, boardmask,
                                                        dev.c_str()));
       readout_cfg = pftool::State::CFG_HCALOPTO_BW;
-      pftool::root()->hide(ONLY_FIBERLESS);
+      pftool::root()->hide(ONLY_FIBERLESS | ONLY_ZCU);
 #else
       pflib_log(fatal) << "Target type '" << target_type << "' requires Rogue.";
       return 1;
@@ -382,7 +382,7 @@ int main(int argc, char* argv[]) {
       auto rocmask = target.get<int>("rocmask", 0x3f);
       tgt.reset(pflib::makeTargetEcalSMMBittware(ilink, rocmask, dev.c_str()));
       readout_cfg = pftool::State::CFG_ECALOPTO_BW;
-      pftool::root()->hide(ONLY_FIBERLESS | ONLY_HCAL);
+      pftool::root()->hide(ONLY_FIBERLESS | ONLY_HCAL | ONLY_ZCU);
 #else
       pflib_log(fatal) << "Target type '" << target_type << "' requires Rogue.";
       return 1;
--- a/app/tool/pftool.h
+++ b/app/tool/pftool.h
@@ -36,6 +36,8 @@ static const unsigned int NEED_FIBER = 0b1u;
 static const unsigned int ONLY_FIBERLESS = 0b10u;
 /// menus/commands that only make sense for the Hcal
 static const unsigned int ONLY_HCAL = 0b100u;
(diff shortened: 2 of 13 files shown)


tomeichlersmith:
Copied from the schematic using the same indexing scheme we did with the ECON-D (`*` is a placeholder for `TRIG` or `DAQ` in the schematic).
- S1_*L -> ROC 0
- S1_*R -> ROC 1
- S2_*L -> ROC 2
- S2_*R -> ROC 3
- S3_*L -> ROC 4
- S3_*R -> ROC 5

ROC | TRG | ECON-T etX
---|---|---
0 | 0 | 6
0 | 1 | 4
1 | 0 | 5
1 | 1 | 7
2 | 0 | 2
2 | 1 | 0
3 | 0 | 1
3 | 1 | 3
4 | 0 | 11
4 | 1 | 9
5 | 0 | 8
5 | 1 | 10

PR #425, opened 2026-06-11
INV_VREF_LUND Update

knh3dg:
- Allows INV_VREF_LUND to update multiple ROCs
- Allows INV_VREF_LUND to update noinv_vref of each link separately
- Changes the linear fitting algorithm to a least squares linear fit
- Removes points with large derivatives to further filter noise
- Renames INV_VREF_LUND to GLOBAL_PEDESTAL_LEVEL
- Renames LEVEL_PEDESTALS to LOCAL_PEDESTAL_LEVEL


app/tool/algorithm/global_pedestal_level.cxx
```diff
@@ -117,109 +117,132 @@ int DataFitter::fit(int target) {
...
+  std::map<int, std::vector<int>> inv_vrefs;
 
   int step = 20;
-
-  while (true) {
-    get_param(tgt, roc, buffer, nevents, step, pedestals_l0, stds_l0,
-              pedestals_l1, stds_l1, inv_vrefs, noinv_vref);
+  bool keep_running = true;
+  while (keep_running) {
+    keep_running = false;
```
tomeichlersmith, line 263: What if ROC0 should be done but ROC1 needs to keep going? What happens to the parameters on ROC0?

I just want to make sure we don't get into a confusing situation where ROC0 successfully levels when its alone but then gets messed up when attempting to level ROC0 and ROC1 at the same time. (ROC0 and ROC1 are just example names for two different ROCs.)

tomeichlersmith (commented):
This makes sense to me and is what I would do as well. My only question is a logical one in regards to how this behaves with multiple ROCs that may converge to settings at different times.

knh3dg:
Now, each link of each ROC should update its noinv_vref separately. I also redid the linear fit. The old version just found the median intercept and slope, which is not always consistent. I switched it for a standard linear fit, and that improved consistency quite a bit. I also added in a filter to remove points with large derivatives, as those arise from noise and were sometimes messing up the selection of the linear region.

cmantill:
May I suggest that the scan is simply renamed as inv_vref_scan (instead of lund?).

tomeichlersmith:
I'm okay with deleting the old `inv_vref_scan`, I don't think anyone uses it. I'm also open to other names that are easier to understand than just the name of the parameter. Maybe `global_pedestal_level` (and then the current `level_pedestals` can become `local_pedestal_level`)?

verenamh:
I ran this updated version here in Lund with one board connected, and it ran without any errors, so it seems to work fine on the fiberless setup as well.

cmantill:
`global_pedestal_level` and `local_pedestal_level` sound great to me!

knh3dg:
I will make that change!

tomeichlersmith (approved):

PR #428, opened 2026-06-26
Updated toa_vref_scan for multiple rocs

### LINKED ISSUE #418: generalize TOA tuning
The current TOA scan and set TOA tasks were developed on the Fiberless setup, but they need to be refactored to use the `SingleECONDRocErxMapping`. This would enable them to be applied to Fiberfull setups. At the same time, these tasks should consider doing all of the active ROCs "at once" i.e. when tuning channel 17 (for example), tune channel 17 for all of the ROCs that are active. This will complicate the code, but will hopefully keep the time of the tuning procedure relatively constant as we scale up from 1 to 6 ROCs.

haseckew:
Updated toa_vref_scan and get_toa_efficiencies to run for multiple rocs (tested on all 4 successfully)


app/tool/algorithm/toa_vref_scan.cxx
```diff
@@ -19,58 +19,125 @@ std::map<std::string, std::map<std::string, uint64_t>> toa_vref_scan(
...
+                          << (int)last_erx << ", eCh " << (int)last_ch;
+        }
+      }
+
+      catch (const std::exception& e) {
+        pflib_log(error) << "[DEBUG MAP] Critical: Mapping lookup threw an "
+                            "exception for ROC "
+                         << i_roc << ". Message: " << e.what();
+      }
+      //
```
tomeichlersmith: What is this doing here? It doesn't seem to be doing anything...

app/tool/algorithm/trim_toa_scan.cxx
tomeichlersmith, line 119: This will scale really poorly as we add more ROCs since this is requiring 100 events for each ROC separately (one Ecal module has 6), but the HGCROCs are pretty well isolated, so we should be able to do a charge injection in one channel on all of the connected ROCs instead of only doing one ROC at a time. (i.e. do ch0 on all ROCs, ch1 on all ROCs, ch2 on all ROCs, etc...). Does that make sense?

app/tool/tasks/trim_toa_scan.cxx
```diff
@@ -8,7 +8,7 @@
 
 void trim_toa_scan(Target* tgt) {
   auto roc{tgt->roc(pftool::state.iroc)};
-  auto settings = pflib::algorithm::trim_toa_scan(tgt, roc);
+  auto settings = pflib::algorithm::trim_toa_scan(tgt, roc, pftool::state.iroc);
```
tomeichlersmith, line 11: So while the toa_vref is processing all of the ROCs, this is telling me that the trim_toa_scan only processes one ROC at a time (and you only tested one of them?). See other comment about trying to do all of the ROCs in "parallel".

tomeichlersmith (changes requested):
I have some feedback on the algorithm and removing some deadcode.

More generally, I'm going to need more evidence that this is working. Maybe some before/after comparisons of the HGCROC parameters? Even better, some before/after plots of the TOA as a function of channel and calib pulse height. I can help guide you on how to make these comparisons if you want. What did you run that made you say it was "tested ... successfully"? These extra details are crucial for making sure the code is working as intended.

haseckew:
CC522399-170E-4EE5-ABBA-30DDB64E5E42_1_105_c
Plotted TOA vref vs efficiency with data aligning with toa_vref_scan's formula of highest_non_zero_eff + 10

ana/toa/toa_efficiency_plot.png
tomeichlersmith, line 1: Please remove this from the source tree

ana/toa/output.yaml
```diff
@@ -0,0 +1,4 @@
+REFERENCEVOLTAGE_0:
```
tomeichlersmith: Please remove this, I don't see a reason for these specific values to be shared across different set ups.

tomeichlersmith: Also, this is only two halves of one ROC. Is the `ana/toa/toa_efficiency.py` plotter only looking at one ROC or do you only have one ROC enabled? Since the whole point of this is to update the toa_vref_scan task to support multiple ROCs, I'd like to confirm that multiple-ROCs can indeed have their toa_vref values determined all from the same data collection runs.

app/tool/algorithm/toa_vref_scan.cxx
```diff
@@ -1,76 +1,164 @@
...
   tgt->setup_run(1, Target::DaqFormat::ECOND_SW_HEADERS, 1);
 
-  std::array<int, 2> target;
-  std::array<std::array<double, 256>, 2> final_effs;
+  std::map<int, std::array<int, 2>> target;  // i_roc int for each variable
+  std::map<int, std::array<std::array<double, 256>, 2>> final_effs;
+  std::map<int, std::map<std::string, std::map<std::string, uint64_t>>>
+      settings;
+
   // TODO 348
```
tomeichlersmith:
```suggestion
```

This comment is to remind us to use the `getRocErxMapping` like you do below.

app/tool/algorithm/trim_toa_scan.cxx
tomeichlersmith: Did you resolve this because it doesn't apply (and we should do it later)? I don't think this is resolved.

tomeichlersmith (changes requested):
Good progress - I am still concerned about if this is actually determining the toa parameters (toa_vref and/or trim_toa) for all of the active ROCs or if this is just doing them for the currently selected `iroc`. Maybe share the terminal output of the toa_vref_scan while printing out the settings ("View deduced settings")?

You also "resolved" some of my comments without changing the code or responding. If you are not going to be able to update the code to address my comments please respond to them saying why you aren't going to do those updates at this time.

tomeichlersmith:
I am merging `main` into this branch to avoid conflict with the auto-formatting applied to https://github.com/LDMX-Software/pflib/pull/431

haseckew:
Screenshot 2026-07-07 at 10 32 20 AM
Screenshot 2026-07-07 at 10 32 34 AM
two methods of obtaining toa_vref values; 1) traditional method, by scanning 255-0 vref and then finding the highest_non_zero efficiency value and then 2) faster method, scans starting at 255 and going down until it finds the first non-zero efficiency value for each link in each roc, then setting that as the toa_vref value;

both of these methods yield identical results, with option 1 being required to plot the data distribution and option 2 being a much more efficient way to obtain data

haseckew:
toa_efficiency_plot_roc0
toa_efficiency_plot_roc1
toa_efficiency_plot_roc2
toa_efficiency_plot_roc3
plots for toa_vref vs efficiency for each roc from this run (in order rocs 0, 1, 2, 3)

ana/toa/toa_efficiency.py
```diff
@@ -20,6 +20,7 @@
     description='takes a csv from tasks.toa_vref_scan and outputs a plot of TOA efficiency per channel against TOA VREF values'
 )
 parser.add_argument('-f', required = True, help='csv file containing scan from tasks.toa_vref_scan')
+parser.add_argument('-r', '--roc', type = int, choices = [0, 1, 2, 3], required = True, help = 'ROC index to filter data (e.g., 0, 1, 2, 3)')
```
tomeichlersmith: The EcalSMM can have up to 6 ROCs, so this should allow the indices 4 and 5 as well.

ana/toa/toa_efficiency.py
```diff
@@ -30,6 +31,11 @@
     sys.exit()
 data, head = read_pflib_csv(args.f)
 
+data = data.iloc[args.roc::4].copy()
```
tomeichlersmith: This selection only works when you _know_ there are 4 active ROCs and worse it fails this assumption silently!

Please include the `i_roc` in the CSV that's written by `toa_vref_scan` so that you can just select on it after loading the data into a dataframe (e.g. with `data[data.roc == args.roc]`). Then if you try to plot roc 2 but roc 2 data was not collected, then your plot will be empty.

app/tool/algorithm/toa_vref_scan.cxx
```diff
@@ -1,74 +1,212 @@
...
+  size_t total_links_to_find = tgt->roc_ids().size() * 2;
+  size_t links_found_count = 0;
 
-  // loop over runs, from toa_vref = 0 to = 255
-  for (int toa_vref{0}; toa_vref < 256; toa_vref++) {
+  DecodeAndBuffer buffer{n_events, tgt->nrocs() * 2};
+
+  // create a .csv file to save efficiency and vref data for analysis
+  std::ofstream csv_file("toa_vref_scan_data.csv");
+  csv_file << "TOA_VREF";
```
tomeichlersmith:
```suggestion
  csv_file << "TOA_VREF,ROC";
```

app/tool/algorithm/toa_vref_scan.cxx
```diff
@@ -1,74 +1,212 @@
...
+    for (int i_roc : tgt->roc_ids()) {
+      const auto& mapping = tgt->getRocErxMapping();
+
+      auto efficiencies =
+          get_toa_efficiencies(i_roc, mapping, buffer.get_buffer());
+      pflib_log(trace) << "got channel efficiencies for ROC " << i_roc
+                       << ", getting max efficiency per link";
+
+      // Save raw channel efficiencies to .csv file
+      csv_file << toa_vref;
```
tomeichlersmith:
```suggestion
      csv_file << toa_vref << ',' << i_roc;
```

app/tool/algorithm/trim_toa_scan.cxx
tomeichlersmith: moving to separate issue https://github.com/LDMX-Software/pflib/issues/430

app/tool/algorithm/toa_vref_scan.cxx
```diff
@@ -1,74 +1,212 @@
...
+  //          pflib_log(info) << "TRIM_TOA for channel " << i << ": " <<
+  //          it->second;
+  //      } else {
+  //          pflib_log(info) << "TRIM_TOA not set for channel " << i;
+  //      }
+  //    }
+  //  } else {
+  //      pflib_log(error) << "No ROC IDs found.";
+  //  }
+
```
tomeichlersmith: While I'm certain this code was helpful for your own development and learning, I think it should be removed if its just going to be commented out.

```suggestion
```

tomeichlersmith (changes requested):

app/tool/tasks/trim_toa_scan.cxx
tomeichlersmith: Work done by @vbe9cs in #439

app/tool/algorithm/toa_vref_scan.cxx
```diff
@@ -1,77 +1,237 @@
...
+#include <fstream>
+
 #include "../daq_run.h"
 #include "../tasks/toa_vref_scan.h"
 #include "get_toa_efficiencies.h"
 #include "pflib/utility/efficiency.h"
 #include "pflib/utility/string_format.h"
+#include "trim_toa_scan.h"
+<<<<<<< Updated upstream
+=======
```
tomeichlersmith: These are leftover from a merge conflict that you did not resolve. If `git` says `CONFLICT`, that means you will need to go through the files it lists and choose which updates to use.

haseckew: this should now be fixed

ana/toa/toa_efficiency.py
haseckew: working on this now, will have this by tomorrow

app/tool/algorithm/toa_vref_scan.cxx
```diff
@@ -1,74 +1,183 @@
...
+     << std::put_time(std::localtime(&in_time_t), "%Y%m%d_%H%M%S") << ".csv";
+  std::string filename = ss.str();
+
+  std::ofstream csv_file(filename);
+  pflib_log(info) << "Saving scan data to file: " << filename;
+  csv_file << "TOA_VREF,ROC";
+  for (int chan = 0; chan < 72; ++chan) {
+    csv_file << "," << chan;
+  }
+  csv_file << "\n";
```
tomeichlersmith: Have the filepath be given as an argument to this function. Then the `task` that calls this function can have the user put in their desired output file (use `pftool::readline_path`). An empty string can signal that the user doesn't want the CSV file written at all (which will be helpful in the long term when we are tuning many chips and have more trust that the algorithm is working as intended).

tomeichlersmith (changes requested):
Looking good :) glad to hear you are working on the plotting. I found that the `i_roc` index was not being written out in the current version of the code so that might need to be patched before the plotting can work.

I'd also request that the writing of the CSV file is optional or I can at least configure where the output file is.

app/tool/algorithm/toa_vref_scan.cxx
tomeichlersmith: We can also achieve the "not writing out" by always writing out to some `ofstream` and then using the special file "/dev/null" to throw away data when we don't want it.

haseckew:
several things have been updated with this most recent commit:

**1. added a yes/no option before running toa_vref_scan on whether or not to write the data to a csv file (for plotting purposes):** 

proof of asking to write all values to csv file (edited csv file name)
proof of asking y:n to write to csv file (said n)
proof of asking y:n to write to csv file (said y)'

_these 3 pictures show that each of the options work and achieve desired results (when saying no to writing to a csv file, there is no print statement showing that it is not writing to the file, as well as it not outputting one after running the scan); the point of this is for efficiency and not writing unnecessary files later down the road._

**2. updated the methodology of the toa_vref_scan to be more efficient:**

    _- one method (the original) scans all 256 reference voltage values and then determines the "highest_non_zero_eff" value and adds 10 and sets that as the reference voltage for that link:_ 

successful results for scanning all values

   _- second method (the updated) starts scanning at vref = 255 and loops down until finding the first non-zero efficiency value for each link in each active roc (and then adding 10 to set the reference voltage), thus saving almost 50% of the runtime as it doesn't need to loop all the way to 0:_

successful results for not scanning all values

**3. as long as a csv file is written for this scan, the data can be plotted:**

   _- these 2 pictures are for when rocs 0 and 1 were scanned and the accompanying data was plotted (note this is the typical distribution found across all tests):_

successful run (all values) roc 0
successful run (all values) roc 1

   _- these 2 pictures are again for rocs 0 and 1 but when the option to not scan all values was selected (the updated scan), these plots are less useful but it is still worth noting that the python file has functionality for both options:_
 
successful run (not all values) roc 0
successful run (not all values) roc 1

    - this picture just confirms that the python file will return an error message if you try to plot data for a roc that wasn't scanned
  
proof of successful plotting (and unsuccessful for roc 2)

tomeichlersmith (approved):
Excellent work 💯 thank you for all of these updates, this looks great!

PR #429, opened 2026-07-01
Update TOT Scan and Add Phase Parameter Scan

### LINKED ISSUE #419: generalize TOT tuning
pretty much identical comments to #418 , just making a separate issue since the algorithm and logic will need to be tested separately.

knh3dg:
- Added a new tot scan which aligns the tot threshold to a set adc target. Examples graphs are below.
- Added a scan which prints information about central_charge_to_l1a, phase_strobe, and phase_ck
- Fixed a type assignment bug in global_pedestal_level and strengthened slope cuts
- Updated channel_wise_calib_scan to function with multiple ROCs enabled (still only scans one roc)
- Updated default settings of local_pedestal_level to be in-line with other commands
adc_linearity_1
tot_s_curve_1


app/tool/algorithm/global_pedestal_level.cxx
```diff
@@ -27,7 +27,7 @@ void DataFitter::sort_and_append(std::vector<int>& inv_vrefs,
   std::vector<DerivPoint> slope_points;
 
   double flat_threshold = 0.1;
-  double linear_threshold = 10;
+  double linear_threshold = 5;
```
tomeichlersmith, line 32: Can you add a comment about how this value was chosen? Just to help future developers understand where this magic number came from.

app/tool/algorithm/tot_scan.cxx
```diff
@@ -0,0 +1,218 @@
+#include "tot_scan_uva.h"
+
+#include "../daq_run.h"
+#include "../tasks/tot_scan_uva.h"
+#include "pflib/utility/median.h"
+#include "pflib/utility/string_format.h"
+
+namespace pflib::algorithm {
```
tomeichlersmith, line 8: in all honesty, I really want to make sure all of us are running the same code. With this in mind, I would like for this to **replace** the prior tot scan and we should ask other folks (namely Lund) to try it out.

app/tool/algorithm/tot_scan.cxx
```diff
@@ -0,0 +1,218 @@
...
+  for (int ch{0}; ch < 72; ch++) {
+    auto ch_page = pflib::utility::string_format("CH_%d", ch);
+    parameters[ch_page]["TRIM_TOT"] = trim_tot_initial;
+  }
+  auto test_params = tgt->tempApplyAllROCs(parameters);
+  // loops through calib, finding calib just before tot activation
+  for (int i_roc : tgt->roc_ids()) {
+    pflib_log(info) << "Getting calibs for roc: " << i_roc;
+    calibs[i_roc].resize(72);
+    for (int ch{0}; ch < 72; ch++) {
```
tomeichlersmith, line 50: So, these two nested for-loops will require a minimum of NROCS $\times$ 72 runs to collect all of the necessary data; however, the separate ROCs are pretty independent, so we can do each channel on the separate ROCs all at the same time. (Making the minium 72 runs instead of NROCS $\times$ 72). Do you think you can implement this rough parallelization?

app/tool/tasks/examine_phase.cxx
```diff
@@ -0,0 +1,165 @@
...
+        }
+      }
+      pflib_log(info) << "phase_ck: " << phase_ck
+                      << " adc: " << pflib::utility::mean(adcs[i_roc]);
+    }
+  }
+}
+
+int peak_bx(Target* tgt, pflib::ROC& roc, int i_roc, DecodeAndBuffer& buffer,
+            int nevents) {
```
tomeichlersmith, line 96: Aligning the peak BX is very helpful, but I want to remind you that we should be running while collecting multiple samples per event (`daq.setup.config`) in order to avoid small timing offsets (common on these test stands) causing data loss.

app/tool/tasks/examine_phase.h
```diff
@@ -0,0 +1,20 @@
...
+#pragma once
+
+#include "../daq_run.h"
+#include "../pftool.h"
+
+/**
+ * TASKS.EXAMINE_PHASE
+ *
+ * Scans over charge_to_l1a and finds bx location of charge pulse.
+ * Scans over phase_ck and prints values. Optimal phase_ck is when pedestal adc
```
tomeichlersmith, line 11:
```suggestion
 * Scans over phase_ck and prints values. Optimal phase_ck is when adc
```
ADC is not a pedestal anymore when we are observing a signal pulse.

tomeichlersmith (changes requested):
Things are looking good and I'm happy to see the plots that look like things are working. Good work!

The biggest request is to try to roughly parallelize the charge injections across the different ROCs but still only pulsing into one channel per ROC.

app/tool/tasks/examine_phase.h
knh3dg: This actually is the pedestal. The idea behind the scan was taken from the attached document page 30. 

[HGCalDPGRunTypes_Jun5.pdf](https://github.com/user-attachments/files/29562282/HGCalDPGRunTypes_Jun5.pdf)

app/tool/algorithm/tot_scan.cxx
knh3dg: Cristina wants me to move on for now, but I can make this change in the future!

app/tool/algorithm/tot_scan.cxx
knh3dg: Will do!

app/tool/algorithm/global_pedestal_level.cxx
knh3dg: Will do!

app/tool/algorithm/tot_scan.cxx
tomeichlersmith: That makes sense, one step at a time

app/tool/tasks/examine_phase.h
tomeichlersmith: cool, thank you for the correct and extra resource :100: i'll add it to our list of resources in the pflib/wiki

tomeichlersmith (commented):
This is looking good, thank you. I am going to use this as a way to have another student at UMN get up to speed on running things with our test stand.

Side note: with a PR open, we have configured GitHub to push formatting commits to your branch, so get in the habit of `git pull` after ~10min you `git push` to a branch with an open PR.

tomeichlersmith:
I am merging main into this branch to avoid conflict with the auto-formatting applied to https://github.com/LDMX-Software/pflib/pull/431

tomeichlersmith:
superseded by #434

PR #434, opened 2026-07-13
Update TOA and TOT scans

### LINKED ISSUE #419: generalize TOT tuning

### LINKED ISSUE #430: parallelize ToT vref/trim scanning
We can only inject charge into one channel **per ROC** at a time, but - when we have multiple ROCs connected to a `Target` - we can do one channel on **each ROC** during a data collection run. This would reduce the number of total data collection runs (`daq_run`) we'd need from 72 $\times$ NROCS to just 72.

As pointed out in https://github.com/LDMX-Software/pflib/pull/429#discussion_r3507816938 this can complicate the data collection code and so was pushed off to get something working first.

These comments also apply to the ToA trim setting where low-amplitude charge injections are used.

knh3dg:
- Added a new toa scan which aligns the toa threshold to a set adc target in the low range. Functions on multiple ROCs in parallel. Example graphs are below.
- Added a new tot scan which aligns the tot threshold to a set adc target in the high range. Functions on multiple ROCs in parallel. Examples graphs are below.
- Added a scan which prints information about central_charge_to_l1a, phase_strobe, and phase_ck
- Fixed a type assignment bug in global_pedestal_level, strengthened slope cuts, and changed the target adc to 150
- Updated channel_wise_calib_scan to function with multiple ROCs enabled, scanning over rocs in parallel
- Updated default settings of local_pedestal_level to be in-line with other commands
- Renamed a previous command labeled "toa_scan" to keep name consistency 

After setting up the ECON/ROC/DAQ, ROC calibration should go as follows:
TASKS.LOCAL_PEDESTAL_LEVEL
TASKS.GLOBAL_PEDESTAL_LEVEL
TASKS.TOA_SCAN
TASKS.TOT_SCAN

If TOA_SCAN or TOT_SCAN is having trouble, e.g. many trim values are 63, try setting the target calib to a higher value. Note that TOA_SCAN aligns in the low range while TOT_SCAN aligns in the high range. Using more events, e.g. 100 instead of 10, should also improve consistency, but from my testing the difference is often quite small.

adc_linearity_roc1
toa_s_curve_roc1
tot_s_curve_roc1


tomeichlersmith (commented):
At first glance, this looks really good; however, it is _a lot_ of new code so I want to test it more thoroughly.

While I'm working on that, how do these developments relate to #433 and #428 ? You delete files that they are editing in those PRs, so does this PR replace those?

knh3dg:
> While I'm working on that, how do these developments relate to #433 and #428 ? You delete files that they are editing in those PRs, so does this PR replace those?

I am pretty sure I have left their files untouched. I did delete some files related to the old tot scan, but I don't think I deleted any files related to the toa scan. Which files are you referring to?

tomeichlersmith:
Oh yea, you're right, I misread `tot_vref_scan` and `toa_vref_scan`.

My question remains though - it looks like you have implemented `toa_scan` which does the tuning of the toa_vref and trim_toa parameters which #433 and #428 are also looking at working. Is the purpose of this function the same or am I misunderstanding?

knh3dg:
Oh yeah, the purpose of the function is the same. The plan was to get a working version. Once Pierce and Will finish their functions and we see how effective they are, we can pick one going forward.

cmantill:
Jack, Pierce and Will summarized progress here and shared it with CMS folks: https://docs.google.com/presentation/d/1PuzthVSWL0KrQL0XiZCgzaWRW_tTlfmDLGLtzZWkPWk/edit?usp=sharing

I'll note that Jack's procedure for the TOA vref and trim tuning is not the same as the one from CMS, whereas Will's + Pierce's follows the CMS procedure. However, Jack's procedure seems to work fine. I suggest merging his algorithms for now. 

- @knh3dg can you add proper documentation in this PR and later on the wiki on the steps you take to run them? 
- when Pierce's + Will's updates are ready we can go back and make a decision on which to keep.

tomeichlersmith (commented):
A second, more thoughtful read of the code (still haven't tried to run it!) gives me more confidence. It looks like a thoughtful algorithm, well done!

You can also add doxygen-style comments within the functions themselves, e.g.
```cpp
  /**
   * First, we incrementally lower toa_vref from its maximum of 255 until XXX
   */
```
This would then end up in the online doxygen manual: https://ldmx-software.github.io/pflib/
(although many of the pages there have gone stale and I need to update them)

tomeichlersmith:
@knh3dg are you waiting for anything from me? I've been waiting for more documentation on your updated algorithms. I'm willing to merge this and have both algorithms available in one version of pflib so they are easier to compare.

cmantill:
I think Jack edited the PR description:
```
TASKS.LOCAL_PEDESTAL_LEVEL
TASKS.GLOBAL_PEDESTAL_LEVEL
TASKS.TOA_SCAN
TASKS.TOT_SCAN
```
If you need more, maybe you can make some explicit requests?

app/tool/tasks/toa_csv_scan.cxx
```diff
@@ -0,0 +1,65 @@
...
+        }
+        f << "\n";
+      },
+      [&](std::ofstream& f,
+          const pflib::packing::MultiSampleECONDEventPacket& ep) {
+        f << calib;
+        // Write the TOA values for each channel
+        for (int ch{0}; ch < 72; ch++) {
+          // TODO 348
+          f << "," << ep.soi().channel(ch / 36, ch % 36).toa();
```
tomeichlersmith: Needs to use the "mapping" in order to deduce the i_erx/i_ch to pass into `channel`. As written, this will swap the upper and lower halves of the ROC for ROC0 on the HcalBackplane (for example).

tomeichlersmith (commented):
Yea, that's probably fine, but as I looked through I noticed one more bug. The toa_csv_scan as written does not "unshuffle" the links appropriately. It needs to use the "mapping" stored within the target to map a roc/channel into a i_erx/i_ch.

app/tool/tasks/toa_csv_scan.cxx
knh3dg: Oh yeah, I did not write this program. I just renamed it (it used to be named toa_scan). I am not sure if it is really used anymore? Maybe we should just remove it.

tomeichlersmith: I'm fine with deleting it

tomeichlersmith (approved):

PR #439, opened 2026-07-14
Trim toa final

### LINKED ISSUE #418: generalize TOA tuning

### LINKED ISSUE #430: parallelize ToT vref/trim scanning

vbe9cs:
Updated TRIM_TOA_SCAN to:
- Fixed issue in the intercepts calculation of the siegel_regression function. Now should use a normal linear regression to find the predicted TRIM_TOA value for our target calib.
- Reorganized the loops in this command so that the function loops in the following order:
```
ch
--trim_toa
----CALIB
------i_roc
```
- Also now can function for multiple ROCs being attached.
- Updated to stop after the first event that is found for each channel on each ROC that triggers this "high efficiency threshold" (we want to see when the toa starts firing very actively, not just a fluctuation). Also this is all the data that we are using anyways so it is not causing the regression to be worse at all.
- Changed efficiency threshold to be when > 95% of the runs return a non-zero TOA value, to make the slope more accurate to where TOA is fully firing.
- Added portion that saves acquired toa data to a csv file to be analyzed later (using toa_graph.py)
- Decided to set a target calib value of 75, may change in the future but I have found that 70-110 is around when toa starts to trigger depending on the channel. (NOTE: this value is only applicable to the HIGHRANGE being active)
- Added skip functionality that skips over a channel if it find that the channel had TOA trigger no times over any of my data collection. This channel has the average TRIM_TOA value of channels on its ROC applied to it. If it is truly not firing this should not affect anything negatively, so this is our best guess of what could be applied to it.
- Applied a clamp to the max and min value of TRIM_TOA that is being applied, as to not overload the register with an int that is too large for it to be stored

Added in the toa_graph.py file which can be used with the command in terminal:

```python toa_graph.py toa_efficiencies.csv [i_roc] [ch]```

This file is intended to be good way to visualize if your data is actually following a linear fit, also can be useful to show if your data is not agreeing with the general "sharp S-curve" shape that we are expecting from TOA efficiency as calib increases. PLEASE NOTE: due to the new portion of the code that breaks after the first high efficiency TRIM_TOA value is found it will be visualized as a line of higher dots that then goes back down to 0, this just means that the scan stopped for this value. You can get far more complete understanding of how that channel is behaving if you remove line 173 of trim_toa_scan.cxx and then do a scan, however this will take significantly longer.

image

Added changes to the toa_vref scan to apply a TRIM_TOA value of 63 to so that its TOA_VREF value doesn't cause the collected data to dip back into the pedestal range once a TRIM_TOA value is applied. Also applied changes to the CHANNEL_WIDE_CALIB_SCAN to make it more efficient to run, by removing scanning over PHASE_CX and multiple bx values as this did not seem to affect any of the data created.

### CODE REVIEW

File: app/tool/algorithm/toa_vref_scan.cxx [hunk 1]
```diff
@@ -19,58 +18,138 @@ std::map<std::string, std::map<std::string, uint64_t>> toa_vref_scan(
...
+                          << (int)first_ch << " | Ch 71  -> eRx "
+                          << (int)last_erx << ", eCh " << (int)last_ch;
+        }
+      }
+
+      catch (const std::exception& e) {
+        pflib_log(error) << "[DEBUG MAP] Critical: Mapping lookup threw an "
+                            "exception for ROC "
+                         << i_roc << ". Message: " << e.what();
+      }
```
tomeichlersmith: are you seeing errors from the mapping? I would consider that a separate bug and not something that should be tested here.

```suggestion
```

app/tool/algorithm/toa_vref_scan.cxx
```diff
@@ -19,58 +18,138 @@ std::map<std::string, std::map<std::string, uint64_t>> toa_vref_scan(
...
+        if (toa_vref == (int)final_effs[i_roc][i_link].size() - 1) {
+          pflib_log(trace)
+              << "[DEBUG SEARCH] Starting backwards search for ROC " << i_roc
+              << " Link " << i_link << ". Initial val at max VREF: "
+              << final_effs[i_roc][i_link][toa_vref];
+        }
+
+        if (final_effs[i_roc][i_link][toa_vref] > 0.0) {
+          highest_non_zero_eff =
+              toa_vref + 20;  // need to add 20 since we don't want to overlap
```
tomeichlersmith: I think I remember this being 10 (and not 20) before. Is there a reason you raised it? If you are seeing a large variation of the pedestals between different channels, that just means you need to level the pedestals (`local_pedetal_level`, `global_pedestal_level`, `local_pedestal_level`) _before_ you try to tune the ToA.

app/tool/algorithm/trim_toa_scan.cxx
```diff
@@ -1,204 +1,379 @@
...
 /**
- * Perform a linear regression using the
- * [Theil-Sen
- * estimator](https://en.wikipedia.org/wiki/Theil%E2%80%93Sen_estimator) which
- * is more robust against outliers.
- *
- * We do the naive algorithm, calculating the slope of all pairs of sample
- * points and then finding the median of that slope. The final intercept is the
- * median of the intercepts from all points using the median slope.
- *
```
tomeichlersmith, line 18: Are you not doing these anymore? A quick scan of the logic below makes me think you still are doing a linear regression (but maybe not doing the median of slope samples?)

app/tool/algorithm/trim_toa_scan.cxx
```diff
@@ -1,204 +1,379 @@
...
+              pflib::utility::string_format("REFERENCEVOLTAGE_%d", i_link)};
+          parameters[refvol_page]["CALIB"] = calib;  // apply CALIB to each link
+        }
+        auto test_params = tgt->tempApplyAllROCs(parameters);
+        // pflib_log(info) << "Applied CALIB = " << calib << " on both links";
+        usleep(10);
+        daq_run(tgt, "CHARGE", buffer, n_events, pftool::state.daq_rate);
+        auto data = buffer.get_buffer();  // take the data from this run
+        int length = data.size();
+        for (int i_roc : tgt->roc_ids()) {
```
tomeichlersmith, line 204: In my opinion, if a ROC has a good trim value, then it should _not_ be included in further analysis of later runs (see suggestion below). And then all runs can cease when all of the ROCs are good, but I am curious if that changes which settings are chosen drastically.

```suggestion
        for (int i_roc : tgt->roc_ids()) {
          // skip ROCs that already have good trim values
          if (good_trim_val[i_roc]) continue;
```

app/tool/algorithm/trim_toa_scan.cxx
```diff
@@ -1,204 +1,379 @@
...
+          }
+        }
       }
     }
+
+    std::map<std::string, std::map<std::string, uint64_t>> parameters2;
+    parameters2[ch_str]["HIGHRANGE"] =
+        0;  // Turn off highrange for the given channel after you are done
+            // scanning over it
+    auto test_params2 = tgt->tempApplyAllROCs(parameters2);
```
tomeichlersmith: While I'm happy to see that you are being careful, I want to point out that the `tempApplyAllROCs` mechanic does this "resetting" for you, so this block is not needed.

```suggestion
```

app/tool/algorithm/trim_toa_scan.cxx
```diff
@@ -1,204 +1,379 @@
...
   }
 
   pflib_log(info) << "sample collections done, deducing settings";
 
+  // saving the efficiencies to a csv file to have visualization using the
+  // toa_graph.py command
+  std::ofstream csv_file("toa_efficiencies.csv");
```
tomeichlersmith: pass this filepath in as an argument to this function and skip writing anything out if the filepath is empty

add a `pftool::readline` prompt asking the user if they way to write out the toa efficiencies for plotting to the `app/task`

tomeichlersmith (changes requested):

tomeichlersmith:
I forgot to say that this is good work :) thank you

also, does this *replace* #433 and #428 ? (i.e. should I close those PRs and delete those branches?)

app/tool/algorithm/toa_vref_scan.cxx
vbe9cs: This was just a leftover from when will was testing this, it is fine to remove.

app/tool/algorithm/toa_vref_scan.cxx
vbe9cs: I raised it because I felt that the offset was not enough to raise the TOA_VREF out of the pedestal range and it had been hurting the TRIM_TOA scan a bit, I think its fine both ways.

app/tool/algorithm/trim_toa_scan.cxx
vbe9cs: No, I am pretty sure that all of the Siegel regression stuff is wrong. Once I run it for all 4 ROCs I can show you that the normal linear regression functions a bit better, but here are pictures of what happens when I ran it with the normal siegel regression:
713_ROC0 csv
713_ROC1 csv
713_ROC2 csv
713_ROC3 csv

app/tool/algorithm/trim_toa_scan.cxx
vbe9cs: That is a good suggestion, but I dont think that stops the runs from happening. It just stops the data from being added to the final_data object, meaning it wont really save a lot of time

tomeichlersmith: Correct - even though runs need to keep happening, individual ROCs which are already "good" aren't changed further. It wouldn't be a speed-up, its just keeping (for example) ROC0's trim value set while ROC1 continues to be tuned.

app/tool/algorithm/trim_toa_scan.cxx
tomeichlersmith: Cool, thanks for the updated plots. Happy to have bugs fixed and algorithms improved!

app/tool/algorithm/trim_toa_scan.cxx
vbe9cs: I think that is fair, thanks for the suggestion!

app/tool/algorithm/trim_toa_scan.cxx
vbe9cs: I have updated this locally, will try to push it soon.

app/tool/algorithm/toa_vref_scan.cxx
```diff
@@ -198,7 +198,13 @@ toa_vref_scan(Target* tgt, bool scan_all, bool write_csv,
...
   }
 
-  return settings;
+  for (int i_link{0}; i_link < 2; i_link++) {
+    std::string page{
+        pflib::utility::string_format("REFERENCEVOLTAGE_%d", i_link)};
+    settings[i_roc][page]["TOA_VREF"] = target[i_roc][i_link];
+  }
+}
+return settings;
```
tomeichlersmith: remove this section. It is now incorrect since the toa_vref scan goes over multiple ROCs each with their own links.

```suggestion
  return settings;
```

app/tool/tasks/toa_vref_scan.cxx
```diff
@@ -58,3 +58,4 @@ void toa_vref_scan(Target* tgt) {
     f << out.c_str() << std::endl;
   }
 }
+}
```
tomeichlersmith: I'm guessing this extra curly brace is one of the syntax errors.

ana/trim_toa/toa_graph.py
```diff
@@ -0,0 +1,53 @@
+import os
```
tomeichlersmith, line 1: Please move this file into the `ana/trim_toa` directory. We've been putting a lot of the other plotting scripts in `ana` and there is a `trim_toa` directory there that seems appropriate.

```
git mv build/toa_graph.py ana/trim_toa/
```

tomeichlersmith (changes requested):
After your merge, there are some syntax errors leftover that you need to patch (e.g. the test fails to compile: https://github.com/LDMX-Software/pflib/actions/runs/29757941083/job/88405044333)

I think the actual updates to trim toa are ready, the leftover comments are just cleanup.

app/tool/algorithm/toa_vref_scan.cxx
```diff
@@ -196,8 +196,12 @@ toa_vref_scan(Target* tgt, bool scan_all, bool write_csv,
           pflib::utility::string_format("REFERENCEVOLTAGE_%d", i_link)};
       settings[i_roc][page]["TOA_VREF"] = target[i_roc][i_link];
     }
+    for (int i_link{0}; i_link < 2; i_link++) {
+      std::string page{
+          pflib::utility::string_format("REFERENCEVOLTAGE_%d", i_link)};
+      settings[i_roc][page]["TOA_VREF"] = target[i_roc][i_link];
+    }
```
tomeichlersmith: This is just a copy of the 5 lines above it.

```suggestion
```

app/tool/tasks/toa_vref_scan.cxx
tomeichlersmith:
```suggestion
```

app/tool/algorithm/trim_toa_scan.cxx
```diff
@@ -1,209 +1,408 @@
...
+          }
+        }
       }
     }
+
+    std::map<std::string, std::map<std::string, uint64_t>> parameters2;
+    parameters2[ch_str]["HIGHRANGE"] =
+        0;  // Turn off highrange for the given channel after you are done
+            // scanning over it
+    auto test_params2 = tgt->tempApplyAllROCs(parameters2);
```
tomeichlersmith: This update got lost in the shuffle of the merge conflict resolution.
You do not need to turn off this parameter since the `tempApplyAllROCs` mechanic does that for you. The object returned by that function will "unset" all of the parameters that were set when it goes out of scope and is destructed.

```suggestion
```

app/tool/tasks/toa_vref_scan.cxx
```diff
@@ -45,16 +45,16 @@ void toa_vref_scan(Target* tgt) {
...
-    std::ofstream f{fname};
-    if (not f.is_open()) {
-      PFEXCEPTION_RAISE("File", "Unable to open file " + fname + ".");
+      std::ofstream f{fname};
+      if (not f.is_open()) {
+        PFEXCEPTION_RAISE("File", "Unable to open file " + fname + ".");
+      }
+      f << out.c_str() << std::endl;
     }
-    f << out.c_str() << std::endl;
```
tomeichlersmith, line 58: I'm not going to merge this change since I don't think it makes sense. It looks like you only allow for the possiblity to save the settings to a file if you also apply them to the chip.

tomeichlersmith (commented):
ok, it looks like some previous fixes were lost in the merge conflict shuffle. I'm just going to do them myself and then merge this.

tomeichlersmith (approved):

PR #444, opened 2026-07-16
Econt decoding

knh3dg:
- Creates a program which can decode ST4 econt data in .csv format

Example input:
[testOutput.csv](https://github.com/user-attachments/files/30098447/testOutput.csv)

How to use:
```
cd pflib/build
./econt-decoder ./testOutput.csv num_lines decode
```
- num_lines is the number of lines you would like to examine
- decode is 1 to view decoded data and 0 to view raw binary data

For example, type
```
./econt-decoder ./testOutput.csv 3 1
```
to see 3 lines of decoded econt data.


tomeichlersmith:
Closing this since it has gone stale and we have code working on `main`.

PR #448, opened 2026-07-27
bias voltage readback with software cache

### LINKED ISSUE #443: unable to readback bias voltage settings on HcalBackplane
**Describe the bug**
When I `SET_SIPM`, the value changes but then "falls back" to all ones (4095). As far as we can tell from physically probing the device, the bias voltage remains at the changed value (i.e. the setting is appropriately applied), but we are unable to read that value more than once.

```
 > SET_SIPM 
Which board?  [1] 
Which (zero-indexed) channel? (-1 for all)  [1] 
Which DAC value?  [0] 42
 > READ_SIPM 
Which board?  [1] 
Which (zero-indexed) channel? (-1 for all)  [1] 
Channel 1: 42
 > READ_SIPM
Which board?  [1] 
Which (zero-indexed) channel? (-1 for all)  [1] 
Channel 1: 4095
```

**To Reproduce**
1. Running pftool with an hcal backplane target
2. `bias.initialize` the board that is active (ROC1 in the example above)
3. try to `set_sipm` and then `read_sipm` or `status` more than once

**Expected behavior**
I'd really like to be able to read the configuration of the bias voltages on the board reliably.

**Context**
Some discussion of this issue has happened over email between @cmantill , @erik-wallin , Lennart, and others.
At the ESA VST in November, Erik soldered an arduino onto the HGCROC board's I2C pins to interact more directly with the board with the `i2ctransfer` command. With that setup, it was observed that separating the write that requests which address(es) to read and the read that gets the resulting values reproduced this issue while doing the write and read within one I2C transaction did not have this issue. The evidence for this is copied below but I have noticed that it does not include proof that multiple reads can be done after the first one in the "single I2C transaction" method.

<details>

<summary>Lennart Comments on Multi-Step Transactions</summary>

```
It depends on what happens between you send a command and that you read something [on the bus]. Stop or repeat start in I2C lingo.
If you do everything in one go, e.g.

[lennart@umn-zcu102-ldmx ~]$ i2ctransfer -y 23 w3@0x1a 0xb0 0xaa 0x55
[this writes 0xaa 0x55 to one of the LED DACs addressed 0x1a]
[lennart@umn-zcu102-ldmx ~]$ i2ctransfer -y 23 w1@0x1a 0x80 r2
0xaa 0x50
[this writes 0x80 to the address 0x1a, and immediately reads 2 bytes back]

like in my bash script everything is well.

But if you split it into two steps
First write 0x80 to the DAC and then read 2 byte we will get back only the written data. I.e. what we last set the DAC to. [But this is merely a reflection of the write command we just sent, not an actual reading of the DAC]

[lennart@umn-zcu102-ldmx ~]$ i2ctransfer -y 23 w3@0x1a 0xb0 0xaa 0x55
[lennart@umn-zcu102-ldmx ~]$ i2ctransfer -y 23 w1@0x1a 0x80
[lennart@umn-zcu102-ldmx ~]$ i2ctransfer -y 23 r3@0x1a
0xaa 0x55 0x80

It looks right, but it is not what we thought it was. More readings this way only return 0xff.

[lennart@umn-zcu102-ldmx ~]$ i2ctransfer -y 23 w1@0x1a 0x80
[lennart@umn-zcu102-ldmx ~]$ i2ctransfer -y 23 r3@0x1a
0xff 0xff 0x80
```

</details>

These observations drew our attention towards the I2C implementation via the lpGBT which we use on the HcalBackplane.

https://github.com/LDMX-Software/pflib/blob/d36af261bea60afaeeca33eb64ab8c6974db5e7a/src/pflib/lpgbt/I2C.cxx#L25-L42

This implementation very clearly uses the "two steps" of one transaction to write the register we want to read and then a second transaction to read the values.

tomeichlersmith:
This ended up being a bit more involved than I expected since (a) we used to copy `Bias` around a lot and `std::array` does not copy its values around to save time (so I made `Bias` no-copy) and (b) the bias menu of pftool was only for HcalBackplanes so I added an intermediate `Target` which both `HcalBackplane` and `HcalFiberless` inherit from and declares the `bias` function .

The software cache uses `std::optional` so that we can be transparent about when we do know that the channel has been set to some value and when we don't.

### DIFF
--- a/app/tool/bias.cxx
+++ b/app/tool/bias.cxx
@@ -5,8 +5,9 @@
  *
  * Only usable for HcalBackplane type targets.
  */
-#include "pflib/HcalTarget.h"
 #include "pflib/Bias.h"
+
+#include "pflib/HcalTarget.h"
 #include "pftool.h"
 
 ENABLE_LOGGING();
@@ -39,10 +40,12 @@ static void bias(const std::string& cmd, pflib::HcalTarget* pft) {
         "Which (zero-indexed) channel? (-1 for all) ", iboard);
     if (ich == -1) {
       for (int i = 0; i < 16; i++) {
-        std::cout << "Channel " << std::setw(2) << i << ": " << bias.readSiPM(i) << std::endl;
+        std::cout << "Channel " << std::setw(2) << i << ": " << bias.readSiPM(i)
+                  << std::endl;
       }
     } else {
-      std::cout << "Channel " << std::setw(2) << ich << ": " << bias.readSiPM(ich) << std::endl;
+      std::cout << "Channel " << std::setw(2) << ich << ": "
+                << bias.readSiPM(ich) << std::endl;
     }
   }
   if (cmd == "READ_LED") {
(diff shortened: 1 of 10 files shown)


tomeichlersmith:
See the wiki page on [SiPM and LED Biases](https://github.com/LDMX-Software/pflib/wiki/SiPM-and-LED-Biases) for more detail on the numbers that should be used.

PR #450, opened 2026-07-29
update ECON-T decoding with 5E+4M decompression and patch off-by-one error

tomeichlersmith:
I noticed an off-by-one error when studying the unpacking this morning and so I wrote a few tests and manually decoded an ECON-T packet I captured last week. This made me more confident that the off-by-one error should be patched and motivated me to put in the decompression of the 5E+4M encoding.


tomeichlersmith:
The errors are saying `Error: OCI runtime error: crun: unknown version specified` which is an issue with a mismatch between the version of `podman` installed and the version of `crun` installed. This is an issue with GitHub's configuration of the CI runners and so I'm going to plow forward despite the red x's. I did the clang format locally and double checked that it compiles in the zcu configuration.

PR #451, opened 2026-07-31
development while attempting to observe cosmics at UMN

### LINKED ISSUE #445: read out trigger path firmware histograms
In the latest ZCU firmware, Jeremy has included a histogram block that is filled with the STC sums from the ECON-T capture. After properly configuring the ROC pedestals and adc thresholds, this would give us effectively live observation of data collection.

https://github.com/slaclab/ldmx-firmware/blob/hcal_zcu102/firmware/targets/LpGBTZCU/ip_repo/shared/histo_quadbyte.v

Reading out the histograms is kinda clunky so I think we'll need some class that can be pointed at the right area of registers and give `read` and `print` functionality so we can read the current status of the histogram and print out a ascii representation of it.

tomeichlersmith:
Breaking my own branch-per-issue rule in order to try to speed up development. Opening this draft PR to keep track of the issues I'm resolving with this work.

- resolves #445 by refactoring Jeremy's initial draft and including a JSON conversion so they can be dumped and re-loaded easily by a python plotter
- add status message to daq_run that prints every 100 events after the first 100
- allow for in-line, trailing comments in pftool scripts
- remove `.cxx` extension from logging channel name
- refactor `TRIG` menu source into multiple files to ease re-compilation
- add extra flags included in the algo output register
- share time-in settings across `trig.timein`, `trig.self_trig`, and `trig.setup` to (hopefully) make it easier to follow the logic
- only apply default alignment capture in ECON-T/lpGBT alignment if it hasn't been set before. Use a default of 30 (note: UMN is using 41/42 now since moving the backplane to the dark room and using much longer optical fibers)
- add `TRIG::get_debug` callback to get access to various status information
- longer pauses after link resets (seemed to be necessary after going to longer optical fibers)


tomeichlersmith:
With a patched firmware and a well-configured set of chips, I am seeing a reasonable STC _encoded_ sum distribution at the trigger level.

fwhist-dump-2026-07-31-173800

tomeichlersmith:
Still not sure why the minimum is 4 but the rest of the distribution looks reasonable and so I'm going to merge these updates and make a new patch release.

PR #452, opened 2026-07-31
LED Bias update for Backplane setup

vbe9cs:
- Updated the led_bias_scan listed in the FiberlessBias branch so that it now works if you only have a backplane setup.
- Added the plotting command that is also listed in the FiberlessBias branch
- NOTE: when using this please scan over a variety of both SiPM and LED DAC values as for most ranges it will look majority like noise. Also make a dark room for your CMB when taking this data.
- Additionally please double check your LED calibration to L1A offset (at the UVA setup the best value is 16-20.)

channel_2_sipmDAC_3400
Example of good run above.


app/tool/tasks/led_bias_scan.cxx
```diff
@@ -0,0 +1,195 @@
+#include "led_bias_scan.h"
+
+#include <filesystem>
+#include <nlohmann/json.hpp>
+
+#include "../daq_run.h"
+#include "../pftool.h"
+#include "pflib/HcalBackplane.h"
```
tomeichlersmith:
```suggestion
#include "pflib/HcalTarget.h"
```

app/tool/tasks/led_bias_scan.cxx
```diff
@@ -0,0 +1,195 @@
...
+#include "pflib/utility/string_format.h"
+// #include "pflib/zcu/HGCROCBoardFiberless.h"
```
tomeichlersmith:
```suggestion
```

app/tool/tasks/led_bias_scan.cxx
```diff
@@ -0,0 +1,195 @@
...
+      {54, 55, 56, 57}, {58, 59, 60, 61}, {63, 64, 65, 66}, {67, 68, 69, 70}};
+
+  // basically taken straight from the current version of bias.cxx render
+  // function
+  auto hcalbp = dynamic_cast<pflib::HcalBackplane*>(tgt);
+  // auto hcalfl = dynamic_cast<pflib::HcalFiberless*>(tgt);
+  // if (!hcalbp && !hcalfl) {
+  //     PFEXCEPTION_RAISE("ValueError", "led_bias_scan only available for
+  //     HcalBackplane or HcalFiberless targets.");
+  // }
```
tomeichlersmith: I have updated the library to better support both HcalBackplane and HcalFiberless by having them both inherit from an `HcalTarget`. Please follow the updated code in bias render:

https://github.com/LDMX-Software/pflib/blob/968b750d7098cc173630a321912d3c6f1ebabd14/app/tool/bias.cxx#L107-L110

```suggestion
  auto hcalbp = dynamic_cast<pflib::HcalTarget*>(tgt);
  if (!hcalbp) {
    PFEXCEPTION_RAISE("BadTarget", "led_bias_scan only available for Hcal targets");
  }
```

app/tool/tasks/led_bias_scan.cxx
```diff
@@ -0,0 +1,195 @@
...
+      }
+      // reset charge_to_l1a to central value
+      tgt->fc().fc_setup_led(central_charge_to_l1a);
+    }
+  }
+  // reset the biases to zero for all channels
+  for (int i = min_cmb_port; i <= max_cmb_port; i++) {
+    bias.setSiPM(i, 0);
+    bias.setLED(i, 0);
+  }
```
tomeichlersmith, line 196: I would much rather reset them to the values they had before the scan rather the always reset them to zero. My reasoning is that we might determine some set of "sensible defaults" that we load and we don't want to have to re-load those defaults after doing a scan of different values.

The solution to this is to use `getSiPM` and `getLED` before you start the `for` loops to create a `std::map` of "original" values. And then use that map here at the end before closing. Even better would be to construct something like a `TempParameters` that does this reset procedure in its destructor. This is nice because then the resetting will happen even if an exception is thrown or the program terminates early (in most cases).

app/tool/tasks/led_bias_scan.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include "../pftool.h"
+#include "pflib/HcalBackplane.h"
```
tomeichlersmith:
```suggestion
```

tomeichlersmith (changes requested):
First read through looks good. I have not looked at the plotting code yet.

Mostly cleanup in the C++ side with one bigger request related to how the settings get "reset" at the end of processing.

app/tool/tasks/led_bias_scan.cxx
```diff
@@ -0,0 +1,196 @@
+#include "led_bias_scan.h"
+
+#include <filesystem>
+#include <nlohmann/json.hpp>
+
+#include "../daq_run.h"
+#include "../pftool.h"
+#include "pflib/HcalTarget.h"
+#include "pflib/utility/string_format.h"
+
```
tomeichlersmith, line 11:
```suggestion
#include "pflib/Bias.h"
```

tomeichlersmith (approved):

PR #454, opened 2026-08-06
just remove any reference to halfwise pages

### LINKED ISSUE #155: Poking parameters halfwise inconsistency
**Describe the bug**

When looking at how channels behave when we turn them off, i.e. setting CHANNEL_OFF = 1 to individual channels, we noticed an inconsistency with doing it channel-wise and halfwise. 

Doing a pedestal run gave an ADC value of 1023 when a channel was turned off by poking it. We tried for different channels and it gave the same result. Trying Lennart's script for turning off parameters also gave an ADC of 1023. 

When turning off channels using the halfwise method, the ADC values remained around 200, even though the individual channel parameters had CHANNEL_OFF set to 1. Doing halfwise with Lennart's script gave the same result. 

To conclude, the halfwise method sets CHANNEL_OFF = 1 on the respective channels, however it doesn't seem to actually turn off the channel. Getting the same result with pflib and Lennart's script suggests a bug with the ASIC.

**To Reproduce**
1. Compile `pflib` with standard config
2. Poke a channel and setting CHANNEL_OFF=1
3. Take a pedestal run. Notice that one of the ADCs is set to 1023.
4. Poke half of the channels halfwise, setting CHANNEL_OFF=1
5. Take a pedestal run. Notice that none of the ADCs are set to 1023. All are around 200.

**Expected behavior**
We weren't exactly sure what to expect of a pedestal run when the channel was turned off. We first assumed that the ADC would be closer to zero since the channel gets grounded, according to Lennart.

**Terminal Print Out**

Figure of ADC set to 1023 from a pedestal run and a channel-wise poke

Image

**Environment:**

### LINKED ISSUE #344: Don't dump HALFWISE page to .yaml
Will overwrite channel-wise parameters.

tomeichlersmith:
While the HGCROC manual mentions `HALFWISE_0` and `HALFWISE_1` and our register maps match the documented addresses listed there, I am not able to observe understandable behavior from them which makes me think that they are an experimental feature that is still buggy.

For example, when I write `CHANNEL_OFF: 1` to `HALFWISE_1`, I would expect either

1. The halfwise pages are write only and the values written to them are copied onto all of the channels for that half. In this case, I would expect `HALFWISE_1.CHANNEL_OFF` to read back as `0` but `CH_##.CHANNEL_OFF` to read back as `1` for all `##` [36, 71].
2. The halfwise pages are parameters that are overlaid on top of the channel parameters if they are non-zero. In this case, I would expect `HALFWISE_1.CHANNEL_OFF` to read back as `1`, the `CH_##.CHANNEL_OFF` parameters would still read back as `0`, but the channels in that half would behave like they were off.

I don't observe either of these behaviors, so I think avoiding this page is the best strategy moving forward. If we get newer versions of the HGCROC or its documentation that update this understanding, we can re-introduce these short-cut parameters.

This indirectly resolves #155 and indirectly resolves #344 by simply removing the page that was causing confusion.

To be clear, I do observe option (1) above for other parameters that are *not* `CHANNEL_OFF` (like `ADC_PEDESTAL`) which does make sense. Maybe I'll just try to remove `CHANNEL_OFF` from the halfwise page to avoid confusion in the future.


tomeichlersmith:
After furthing investigation, I don't think this should be done. I am instead going to work towards having a special exception for `HALFWISE` when the parameters are dumped and when the "prior registers" are determined.

PR #455, opened 2026-08-07
Iss407 pedestal refactor

duncanswilmot:
Opening a PR but may need to do some extra work before merging.

For local pedestal leveling, change baseline `TRIM_INV` from `0` to `32`, add adaptive support for both `ECAL` and `HCAL` ROCs, and mask dead channels (i.e., those with 0 standard deviation).  Seeing some difficulty leveling for ECAL (see image). May just be running out of range with `TRIM_INV`, but someone should try to run this on their own to verify behavior. 

Note: did not yet test for HCAL

ecal_smm_pedestal_local

### DIFF
--- a/app/tool/algorithm/local_pedestal_level.cxx
+++ b/app/tool/algorithm/local_pedestal_level.cxx
@@ -74,9 +74,9 @@ local_pedestal_level(Target* tgt) {
   // each channel has measurements at different parameter points
   std::map<int, std::array<int, 72>> baseline, highend, lowend;
   // will fill with standard deviations so we can ommit dead channels
-  std::map<int, std::array<int, 72>> basedevs, highdevs, lowdevs;
+  std::map<int, std::array<int, 72>> basedevs;
   // include all zero parmeter run values since hcal used DACB and SIGN_DAC
-  std::map<int, std::array<int, 72>> allzero, zerodevs;
+  std::map<int, std::array<int, 72>> allzero;
 
   DecodeAndBuffer buffer{n_events, tgt->nrocs() * 2};
 
@@ -94,9 +94,7 @@ local_pedestal_level(Target* tgt) {
 	    parameters[ch_str]["TRIM_INV"] = 32; 
 	  }
     }
-    std::cout << "im alive" << std::endl;
     auto test_params = tgt->tempApplyAllROCs(parameters);
-    std::cout << "im still alive" << std::endl;
 
     daq_run(tgt, "PEDESTAL", buffer, n_events, 100);
     pflib_log(trace) << "baseline run done, getting channel medians";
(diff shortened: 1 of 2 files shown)


app/tool/algorithm/local_pedestal_level.cxx
```diff
@@ -38,6 +41,23 @@ static std::array<int, 72> get_adc_medians(
...
+  std::array<int, 72> stdevs;
+  /// reserve a vector of the appropriate size to avoid repeating allocation
+  /// time for all 72 channels
+  std::vector<int> adcs(data.size());
+  for (int ch{0}; ch < 72; ch++) {
+    for (std::size_t i{0}; i < adcs.size(); i++) {
+      auto [i_erx, i_ch] = mapping.toErxChannel(i_roc, ch);
+      adcs[i] = data[i].soi().channel(i_erx, i_ch).adc();
+    }
+    stdevs[ch] = pflib::utility::stdev(adcs);
```
tomeichlersmith, line 56: This is silently converting the `double` into an `int`, probably just truncating off everything after the decimal point. This is a problem because a channel could be quiet and have a noise of 0.9 ADC which then gets truncated to 0 and ignored later in the tuning.

My suggestion is to never do the `sqrt`, just use the variance so that we never convert to double. That way, we can check for dead channels by exact equality with zero.

app/tool/algorithm/local_pedestal_level.cxx
```diff
@@ -61,20 +85,28 @@ local_pedestal_level(Target* tgt) {
...
+      } else {
+        parameters[ch_str]["TRIM_INV"] = 32;
+      }
     }
     auto test_params = tgt->tempApplyAllROCs(parameters);
 
     daq_run(tgt, "PEDESTAL", buffer, n_events, 100);
     pflib_log(trace) << "baseline run done, getting channel medians";
     std::map<int, std::array<int, 72>> medians;
+    std::map<int, std::array<int, 72>> stdevs;
```
tomeichlersmith: While it is probably optimized out, there is no reason to go through this extra `stdevs` local variable. The only reason I have the local `medains` variable is so that we can compute the `target` later (which requires sorting) after copying out the values on a per-channel basis (where we want to keep the original, channel-indexed order).

tomeichlersmith (changes requested):
This is just a first pass - I have not tried to run it myself yet, but the actual algorithm looks good. I just want to avoid one pitfall and drop one extra local variable that I don't believe is necessary.

duncanswilmot:
I couldn't avoid running into the same issue with variance as standard deviation, so I just attempted to keep is as double the whole way rather than adding new code for a variance function. Does this work?

app/tool/algorithm/local_pedestal_level.cxx
tomeichlersmith: As we discussed in person, since the `mean` is already a `double` and a non-broken RMS has a clear range (>0.5 at least), we can leave the stddev as type `double` and change the check from equality to less than.

app/tool/algorithm/local_pedestal_level.cxx
```diff
@@ -131,6 +193,9 @@ local_pedestal_level(Target* tgt) {
       settings;
   for (int i_roc : tgt->roc_ids()) {
     for (int ch{0}; ch < 72; ch++) {
+      if (basedevs[i_roc].at(ch) == 0) {
```
tomeichlersmith:
```suggestion
      if (basedevs[i_roc].at(ch) < 0.1) {
```

tomeichlersmith (commented):
Can you test that this will ignore channels that we turned off?

You can turn off a channel with
```
 > ROC
 > POKE
Page: CH_##
Parameter: CHANNEL_OFF
Value: 1
```
where `##` is the channel number you want to turn off on ROC0.

duncanswilmot:
Now comparing stdev to 0.1 and printing out ignored channels. 
Seem to be losing some extra channels when I turn off CH_0 with `POKE`

* Before turning off CH_0
CH_0_before_poke

* After turning off CH_0
CH_0_after_poke

tomeichlersmith:
I'm not too surprised other channels are getting ignored (they probably correspond to the channels missing from your initial plot), but the additional channels getting ignored after turning off ch_0 is worrying. Can you repeat the test a few times to make sure? Also, do a `daq.pedestal` run in between each test run of pedestal leveling so we can see which channels we expect to be ignored manually.

duncanswilmot:
Reproduced a couple times. The data were wonky on this run, but should still show the channels we expect to be dead. 

ecal_smm_pedestal_test

tomeichlersmith:
Ok, the EcalModule follows this mapping

https://github.com/LDMX-Software/pflib/blob/276c4879ec1df9d779c5c89a56630527ff5b2b03/src/pflib/EcalModule.cxx#L114-L115

ROC2 ends up in the first 2 links with its halves in the correct order, but your code does not report ignoring channels 7 and 9 (but they are absent from the plot of Link 0 implying they have uniform pedestals). Your code reports ignoring channel 12 which is also absent from the Link 0 plot.

This tells me either your code is not working as you are assuming it works or the hardware is unstable and channels are floating into/out of "working". In either case, I think this needs more careful study. I don't really care if the hardware has some channels permanently broken (that's actually helpful for testing this feature), what does matter is that it is changing. If the hardware is actually unstable at this level, that is very bad for our setup at UMN making it difficult to test any kind of EcalSMM development.

PR #457, opened 2026-08-12
Single photon Spectrum event

vbe9cs:
This has been our attempt at UVA to make a scan that is able to take data from a large amount of LED daq_runs (measuring over multiple bunch crossings and phase_ck values).
 - First the SPS_READOUT command takes lots of input values (LED_DAC, SiPM_DAC, BX, etc...) and applies them to the ROC and HDMI port of interest. Next it scans using a LED scan and writes the data to a CSV file
 - This CSV file is then interpreted and made into histograms with the SPS_histogram.py file (See pic below)
image
 - To run this correctly:  before you run the SPS_READOUT command please find a correct TRIM_INV value to set the port to (my stratedgy is to take one of the values found by running LOCAL_PEDESTALS and then using it for all of them, but only really focusing on the channel that the TRIM_INV value corresponds to (Could be updated in the future), this is done to attempt to lower the spread of values). Next you run the SPS_READOUT command. The settings that we have been setting at uva with multiple ROCs attached to our backplane have been:

1. 1000-10000 events per time point
2. 22 as the Target bunch crossing
3. Number of bunch crossing to scan over = 2
4. SiPM_DAC = 3850
5. LED_DAC = 3000
6. Only doing the PHASE_CK value of 2 and no other values
7. TRIM_INV of 9 (it works for one of the channels)
9. Range of TRIM_INV values: 1 (used for DNL later)
10. inputdac = 32 (guess)

 - Once you have taken a run go to the ana/LED directory and run "python3 SPS_histogram.py /full/path/" and it will produce the desired graphs
 - Please note: this is only a start, at UVA we have had very little success actually getting this to work with regards to getting a substantive spectrum for the photons interacting with the SiPM, there are many settings that we have not confirmed are correct. Most of these come from the CSM documentation that we could find and some previous experience in the LED_BIAS_SCAN settings. Any additional work on alternative settings would greatly be appreciated, but I feel that it is focus on what the correct value for inputdac is and what the biasing should be for this scan.
 - NOTE: this is partially incomplete and I have not added a functionality to add all of the data from a single run that are on different bunch crossings together in any way, so I believe that the function just records them as entries in the CSV file, need to update this.

### DIFF
--- a/ana/LED/SPS_histogram.py
+++ b/ana/LED/SPS_histogram.py
@@ -1,3 +1,4 @@
+<<<<<<< Updated upstream
 import numpy as np
 import matplotlib.pyplot as plt
 import pandas as pd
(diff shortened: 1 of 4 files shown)


ana/LED/SPS_histogram.py
```diff
@@ -0,0 +1,61 @@
...
+
+        max_adc = int(np.max(data_hist))
+        min_adc = int(np.min(data_hist))
+        bins = np.arange(min_adc - 0.5, max_adc + 1.5, 1)
+
+        counts, edges = np.histogram(data_hist, bins)
+        errors = np.sqrt(counts)
+        centers = (edges[:-1] + edges[1:]) / 2
+        hep.histplot((counts, edges), ax=ax, histtype='step', density=False, label=labels)
+        ax.errorbar(centers, counts, yerr=errors, fmt='.', capsize=2, markersize=3)
```
tomeichlersmith, line 51: This mechanic of filling histograms and plotting them with error bars is also accomplished by the `hist` package https://hist.readthedocs.io/en/latest/
I would suggest using it because it makes your code simpler and it has sensible defaults for error calculations and plotting styles.

app/tool/tasks/SPS_readout.cxx
```diff
@@ -0,0 +1,163 @@
...
+      {54, 55, 56, 57}, {58, 59, 60, 61}, {63, 64, 65, 66}, {67, 68, 69, 70}};
+
+  tgt->setup_run(1, Target::DaqFormat::ECOND_SW_HEADERS, 1);
+  pflib::DAQ& daq = tgt->daq();
+
+  auto hcalbp = dynamic_cast<pflib::HcalTarget*>(tgt);
+  if (!hcalbp) {
+    PFEXCEPTION_RAISE("BadTarget",
+                      "led_bias_scan only available for Hcal targets");
+  }
```
tomeichlersmith, line 57: Move this to the beginning of the function so nothing is run if its the wrong target type.

app/tool/tasks/tasks.cxx
```diff
@@ -79,7 +80,10 @@ auto menu_tasks =
                "calibrate TRIM_TOA parameters for each channel", trim_toa_scan)
         ->line("TOA_SCAN", "calibrate TRIM_TOA parameters for each channel",
                toa_scan)
-        ->line("LED_BIAS_SCAN", "Sweeps SiPM and LED DACs", led_bias_scan);
+        ->line("LED_BIAS_SCAN", "Sweeps SiPM and LED DACs", led_bias_scan)
+        ->line("SPS_READOUT",
+               "Take data of a photospectrum on one bunch crossing",
+               sps_readout);
```
tomeichlersmith, line 86: we can also prevent `SPS_READOUT` from being display for non-Hcal targets
```suggestion
               sps_readout, ONLY_HCAL);
```

which we should probably add to the `LED_BIAS_SCAN` as well

app/tool/tasks/SPS_readout.h
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * TASKS.SPS_SCAN
+ *
+ * Used to scan different SiPM or LED DAC values/biases.
+ * Both can either be varied over a range, or kept constant if the same value is
+ * entered for the start and stop value. One can choose how many ports have CMBs
+ * connected An LED flash on one CMB flashes into four HGCROC channels
+ * simultaneously, thus the pulses of all four respective channels are recorded
+ * in the csv file.
```
tomeichlersmith, line 13: This is just a copy of led_bias_scan's description? Please update

app/tool/tasks/SPS_readout.cxx
```diff
@@ -0,0 +1,163 @@
...
+  fname = pftool::readline_path("SPS-scan", ".csv");
+
+  int g = 0;
+  int phase_ck = 0;
+
+  DecodeAndWriteToCSV writer{
+      fname,
+      [&](std::ofstream& f) {
+        nlohmann::ordered_json header;
+        f << std::boolalpha << "# " << header << '\n'
```
tomeichlersmith: This JSON header is useful for including extra metadata about the run.
Things I would suggest including:
- ROC/Channel that is being studied
- target bx
- other parameters that are constant for the run but different from the defaults

app/tool/tasks/SPS_readout.cxx
```diff
@@ -0,0 +1,163 @@
...
+        .add(refvol_page, "CALIB_2V5", 0)
+        .add(refvol_page, "INTCTEST", 1)
+        .add(refvol_page, "CHOICE_CINJ", 0)
+        .add(global_page, "CD", 2)
+        .add(global_page, "CF", 8)
+        .add(global_page, "RF", 10)
+        .add(channel_page, "HIGHRANGE", 0)
+        .add(channel_page, "LOWRANGE", 0)
+        .add(calib_page, "INPUTDAC",
+             32)  // No idea what this should be (MAXES out at 63)
```
tomeichlersmith, line 191: Please add more comments describing what these parameters are trying to do. This is more than just "turning on charge injections" as stated in your comment above.

Add comments in the line above the parameter just to make the formatting easier
```suggestion
        // No idea what this should be (maxes out at 63)
        .add(calib_page, "INPUTDAC", 32)
```

app/tool/tasks/SPS_readout.cxx
```diff
@@ -0,0 +1,163 @@
...
+
+      auto phase_test_handle =
+          roc.testParameters().add("TOP", "PHASE_CK", phase_ck).apply();
+
+      bool enable_l1a_follow;
+      // int central_charge_to_l1a = tgt->fc().fc_get_setup_led();
+
+      // tgt->fc().fc_setup_led(start_led_new);
+      tgt->fc().fc_setup_led(tgt_bx);
+      pflib_log(info) << " Target BX  = " << tgt_bx << "\n";
```
tomeichlersmith, line 257: So, you change the led bx offset here but do not revert it back to the value it had before. Please add the "revert" mechanic so the led bx offset is only ever permanently updated when the user intends to update it.

app/tool/tasks/SPS_readout.cxx
```diff
@@ -0,0 +1,163 @@
...
+  auto& bias = hcalbp->bias(iboard);
+  auto& mapping{tgt->getRocErxMapping()};
+
+  int cmb_port = pftool::readline_int("Channel to scan on? ", 0);
+  int nevents = pftool::readline_int("How many events per time point? ", 1000);
+  int start_led = tgt->fc().fc_get_setup_led();
+  int tgt_bx =
+      pftool::readline_int("Target BX? (~22 should be BX = 4) ", start_led);
+  int len_bx = pftool::readline_int("Number of BX to scan over? ", 2);
+  tgt->fc().setL1AperROR(len_bx);
```
tomeichlersmith, line 72: Besides sending a burst of extra L1A per ROR, we also need to inform the data collection side that it should expect multiple samples.

https://github.com/LDMX-Software/pflib/blob/f856638f4c9fa5af020b9cfbf4e9d1fcce833df7/app/tool/daq.cxx#L112-L114

It might be easier to just not allow users to change the number of BX being collected here and instruct them to use `daq.setup.config` if they want to use multi-sampling.

app/tool/tasks/SPS_readout.cxx
```diff
@@ -0,0 +1,163 @@
...
+      daq_run(tgt, "LED", writer, nevents, pftool::state.daq_rate);
+      usleep(10);
+      // auto data = buffer.get_buffer();
+      // auto mapping = tgt->getRocErxMapping();
+      // auto [i_erx, i_ch] = mapping.toErxChannel(i_roc, 17);
+      // for (std::size_t i{0}; i < data.size(); i++) {
+      // for (int j = 0; j < nr_bx; j++) {
+      // adcs[j].push_back(data[i].samples.at(j).channel(i_erx, i_ch).adc());
+      //}
+      //}
```
tomeichlersmith, line 268: Please remove code that is now commented-out and not in use.

```suggestion
```

tomeichlersmith (changes requested):
This looks good! Thank you

I have a lot of nit-picky updates, but nothing too serious that changes the overall logic or intention.

haseckew:
just adding my updates to the SPS script and plotting (more for Cristina's use); they are untested so Tom you don't really need to look at them right now, will merge again once tested

app/tool/tasks/SPS_readout.cxx
```diff
@@ -118,10 +182,39 @@ void sps_readout(Target* tgt) {
...
+      auto cm_page = pflib::utility::string_format("CM_%d", link);
+       test_param_builder.add(cm_page, "GAIN_CONV", 4); //0 to 7
+    }
+    test_param_builder.add(channel_page, "GAIN_CONV", 4); //0 to 7
+>>>>>>> Stashed changes
   }
 
   auto test_param_handle = test_param_builder.apply();
 
+<<<<<<< Updated upstream
```
tomeichlersmith, line 217: These merge conflicts arose because

1. You opened this PR which triggered an auto-formatting commit to be pushed
2. You also edited those same files

The good news is that the auto-formatting can be re-run pretty easily, so if you want to avoid these merge conflicts in the future you can force push your commits and delete the auto-formatting commit. You can also do formatting locally or just remember to `git pull` before you make more changes on the branch.

PR #459, opened 2026-08-14
FW trigpath block refactor

tomeichlersmith:
In order to prepare the trigger path firmware for translating it onto the APx, I moved the lpGBT decoding from the `standardLpGBTPair` block into the `trigpath` block. On the software side, this affects which registers the `ZCUOptoLink` should look at when it is trying to check on the status and rates of the trigger fiber (as opposed to the DAQ fiber).

Leaving this as a draft since I want to clean up the code more and get a nice version of the firmware built that I can post somewhere for others to download and use.

### DIFF
--- a/src/pflib/zcu/zcu_optolink.cxx
+++ b/src/pflib/zcu/zcu_optolink.cxx
@@ -12,7 +12,7 @@ namespace zcu {
 ZCUOptoLink::ZCUOptoLink(const std::string& coder_name, int ilink, bool isdaq)
     : transright_("transceiver_right"),
       coder_(coder_name),
-      coder_name_(coder_name),
+      coder_name_(isdaq ? coder_name : "trigpath-0"),
       ilink_(ilink),
       isdaq_(isdaq),
       the_log_{logging::get("zcu_optolink")} {
@@ -22,7 +22,6 @@ ZCUOptoLink::ZCUOptoLink(const std::string& coder_name, int ilink, bool isdaq)
   transright_.write(0x2, 0xF0000);
   int chipaddr = 0x78;          // EC
   if (isdaq) chipaddr |= 0x04;  // IC
-
   transport_ =
       std::make_unique<lpGBT_ICEC_Simple>(coder_name, !isdaq, chipaddr);
 }
@@ -31,6 +30,7 @@ static const uint32_t REG_STATUS = 3;
 
 void ZCUOptoLink::soft_reset_link() {
   /// reset the decoder for the current link
+  // trigpath update
   coder_.write(0, 1 << (ilink_ % 2));
 }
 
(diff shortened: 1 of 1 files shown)


src/pflib/zcu/zcu_optolink.cxx
```diff
@@ -11,7 +11,8 @@ namespace zcu {
 
 ZCUOptoLink::ZCUOptoLink(const std::string& coder_name, int ilink, bool isdaq)
     : transright_("transceiver_right"),
-      coder_(coder_name),
+      // TODO generalize
+      coder_(isdaq ? coder_name : "trigpath-0"),
```
tomeichlersmith: generalize to include `trigpath-1` and document why this is happening and we aren't just using `coder_name` or having `coder_name` be `trigpath-N`.

src/pflib/zcu/zcu_optolink.cxx
```diff
@@ -173,7 +182,7 @@ std::map<std::string, uint32_t> ZCUOptoLink::opto_rates() {
     for (int i = 0; i < cnames.size(); i++) {
       retval[cnames[i]] = coder_.read(CRATES_OFFSET + i);
     }
-  } else {
+  } else if (isdaq_) {
```
tomeichlersmith, line 219: clean-up and make it appear the same as the trig link

src/pflib/zcu/zcu_optolink.cxx
```diff
@@ -134,15 +135,23 @@ std::map<std::string, uint32_t> ZCUOptoLink::opto_status() {
...
+    retval[prefix + " READY"] = (val >> (0 * 2 + (ilink_ % 2))) & 0x1;
+    // NOT_IN_RESET is resetn_decoder_gth_clock
+    retval[prefix + " NOT_IN_RESET"] = (val >> (1 * 2 + (ilink_ % 2))) & 0x1;
+    /**
+     * The LINK_ERRORS are not being incremented when a known error is injected,
+     * so without a future firmware patch, we are electing to ignore them.
+    retval[prefix + " LINK_ERRORS"] = coder_.read(4 + (ilink_ % 2)) & 0xFFFFFF;
+     */
+  } else {
+    val = coder_.read(0xC04 / 4);
```
tomeichlersmith: I use this constant in a few places, should make it a named variable.

tomeichlersmith:
Notes about software changes from firmware cleanup...

- spam/inject error parameters in lowest 2 bits of register 6
- reset decoder clock in bit 1 of register 2
- decoder ready in bit 0 of register 2
- rates re-compressed, dropping `trig` registers in `standardLpGBT` that are moved to `trigpath`

src/pflib/zcu/zcu_optolink.cxx
tomeichlersmith: also update version printout to include this argument and its version

